PR #21, opened 2017-06-13
Feature/form group

gelsto:
Note: This is only for markup/props/documentation. The actual styling isn't there yet :)
cc @sumul 
This has a lot of documentation in it (woo fractal!), but as an overview:
### `<FormControl />`
Supports a documented list of HTML attributes and synthetic event handlers. Right now I'm only allowing `text`, `email`, and `password` as the `type` attribute. I know we want to support others but I'm not sure we have to do that right now.
### `<FormGroup />`
The 2 basic patterns for using this component are (1) passing the `controlID` prop and having it take care of accessibility or (2) not passing it and dealing with it yourself. I've outlined a few examples for when you'd want to use each one.

I'm interested if this is how we're thinking about doing responsive size variants: 4af1a29. We might not need to make every variant available at every breakpoint?


components/_preview.jsx
```diff
@@ -9,7 +9,11 @@ const ROOT_ELEMENT_ID_PREFIX = 'root-';
 class Variant extends React.Component {
   render() {
     return (
-      <div id={this.props.rootElementId} dangerouslySetInnerHTML={this.renderInner()} />
+      <div
+        id={this.props.rootElementId}
+        dangerouslySetInnerHTML={this.renderInner()}
+        style={{marginBottom: 20}}
```
gelsto: I added this so the variants don't look so squished on preview. Not very elegant, we might want to use something else as we add more components?

NickDunkman: This is what our fancy new spacing classes are for! Can do this:

```jsx
className="mb5"
```

or mb4.

`mb4` = `16px`
`mb5` = `24px`

components/form_control/_form_control.scss
```diff
@@ -0,0 +1,33 @@
+@mixin form-control-sizer($size) {
+  @if $size == 'small' { font-size: $font-size-2; }
+  @if $size == 'large' { font-size: $font-size-5; }
+}
+
+.form-control {
```
NickDunkman: I think we should use BEM for this one, since we have modifiers, and also so it matches the react component (`<FormControl>`). So we'd have classes like this:

```sass
.FormControl
.FormControl--small
.FormControl--large
.FormControl--large-md
etc.
```

components/form_control/_form_control.scss
```diff
@@ -0,0 +1,33 @@
...
+.form-control-small {
+  @include form-control-sizer($size: 'small');
+}
+
+.form-control-large {
+  @include form-control-sizer($size: 'large');
+}
+
+@if $enable-type-utilities {
+  @each $breakpoint in map-keys($grid-breakpoints) {
```
NickDunkman: Yea I think it's probably best to just provide things at all our defined breakpoints, since when we need them it'll be a little tedious to:

1. add the new classes to phenotypes code
2. make a phenotypes pr
3. publish to npm
4. update pare_web to use new phenotypes version
5. ...

components/form_control/_form_control.scss
gelsto: My understanding was that we weren't using BEM for components that only have one element, even if they have modifiers (ex. `.cta-button`). I feel good about changing it, but we might want to make the distinction clearer in the styleguide.

components/form_control/_form_control.scss
gelsto: 5. Profit

gelsto: That makes sense. I was just pointing it out in case we wanted to limit nonsensical responsive classes like `.FormGroup--small-xl`, but it doesn't really matter. We can probably extract some of this logic into a helper if we'll be using it often.

components/_preview.jsx
gelsto: Omg duh. That's embarrassing.

components/form_control/_form_control.scss
gelsto: Also maybe should move the styleguide to the wiki on this repo

NickDunkman: Yea agree on putting styleguide here. Should put a lint-scss here too.

components/_preview.jsx
NickDunkman: Haha. Yea can maybe come up with standard "canvas" components for the fractal examples that use a standard spacing system, but for now can just wing it with those.

components/form_control/_form_control.scss
NickDunkman: The rule in our wiki is actually single _class_ (not elements). I intended to allow a class for one element with some modifier classes to be BEM.

I think we should probably use BEM for all components tho ... still kind of feeling out what the right balance is, but current thought is:

BEM: 

- components (the class describes the component)
- multiple classes (modifiers, inner elements, etc.)
- can safely change these b/c they're only used by those kind of components

non BEM:

- classes describe what the css properties are doing (e.g. "hide" or "mb1")
- often only contain one style property
- used by any type of element/component, so should not be changed except in rare cases (e.g. if you find a a better way to do the thing the class is describing, such as if a new css property comes out)

ALSO:

When designing a component's css classes ... it's best to put the required styles in the BEM classes (all the styles that the component should always have). The optional things can be done using non BEM utility classes. A good example was the `.Meter` in our doctor scorecard .... you can optionally have each bar in the meter be equal width or different percentages, using our `.width-X` classes.

components/form_control/_form_control.scss
NickDunkman: Yea -- if this was in pare_web, I would definitely be in favor of not adding `.FormGroup--small-xl` until we actually need it. But since this is designed to be used by any old project, you want to give the designer the ability to do whatever (such as make a form group that actually gets smaller on a bigger screen)

components/form_control/_form_control.scss
gelsto: `.FormControl` is used by a different component, though, which seems tricky?

NickDunkman: You mean inside of a `.FormGroup`?

gelsto: I'm also confused about the required styles vs. utility classes in this case.  If one of the goals is to allow people to just copy/paste html, shouldn't we not be adding a bunch of font sizing/text coloring/etc. classes on there?

gelsto: yeah:
 ```sass
.FormGroup--has-error {
  .FormControl {
    border-color: $red;
  }
}
```

Comment (NickDunkman) [inline components/form_control/_form_control.scss]: Yea I think that's a code smell -- probably not always 100% avoidable, but should avoid nesting different components in css. Could solve this like this:

```html
<div class="FormGroup FormGroup--has-error">
  <input class="FormControl FormGroup__control" />
</div>
```

```sass
// FormControl.scss owns the presentation logic for different modifications of FormControl

@mixin form-control-error {
  border-color: $red;
}

.FormControl--error {
  @include form-control-error;
}

// FormGroup.scss can use that mixin for convenience

.FormGroup--has-error {
  .FormGroup__control {
    @include form-control-error;
  }

  .FormGroup__select {
    @include select-error;
  }
}

File: components/form_group/_form_group.scss [hunk 4]
```diff
@@ -0,0 +1,63 @@
...
+  font-size: $font-size-1;
+}
+
+.FormGroup--has-error {
+  .FormGroup__control {
+    @include form-control-error-state;
+  }
+}
+
+.FormGroup--small {
```
NickDunkman: I think we should apply the `small` size to the base `FormGroup` class and omit this rule. Can still have responsive classes for the small size.

gelsto: There's a default medium size that doesn't have a modifier

NickDunkman: 👍  ... think generally we should use the smallest size as the default (or more specifically, the size we want to default to on small screens with less hardware, so they have less work to do)

guides/10-component-conventions.md
```diff
@@ -0,0 +1,7 @@
+---
+status: draft
+---
+
+## phSize
```
NickDunkman: Also discussed: given these two philosophies:

1. just dump all the responsive modifiers on the elements and let css adjust using media queries
2. add/remove modifiers using react as screen size changes

The first probably makes more sense for these low level phenotypes components, so that they don't have to subscribe to the screen state, and it's probably less work for the browser

components/form_control/_form_control.scss
```diff
@@ -0,0 +1,46 @@
+// scss-lint:disable SelectorFormat
+@mixin form-control-sizer($size) {
+  @if $size == 'small' { font-size: $font-size-2; }
+  @if $size == 'large' { font-size: $font-size-5; }
+}
+
+@mixin form-control-error-state() {
```
sumul: I'm curious about this mixin and the one below. They only get used in one place, which reduces readability for me, especially since they only have one rule. Are you expecting to build more fancy stuff onto these in the future?

components/form_control/form_control.jsx
```diff
@@ -0,0 +1,88 @@
...
+function ariaProps(props) {
+  return Object.keys(props).reduce((ariaAttrs, key) => {
+    if (key.match('^aria-')) {
+      return Object.assign(ariaAttrs, { [key]: props[key] });
+    }
+    return ariaAttrs;
+  }, {});
+}
+
+function FormControl(props) {
```
sumul: I think we should rename this component to `TextInput`. I took a closer look at Bootstrap's nomenclature, and I don't like how `.form-control` can be applied to text inputs, textareas, and select menus to provide styling while also having semantic meaning within the `.form-group` module. I think it would be cleaner for a `.textinput` (or similar) class to be a single-purpose utility: apply specific styling to text-type inputs. Then we could have `.FormGroup__control` be another single-purpose class that does the expected thing when used within the `.FormGroup` module. IMO, separating these names will make a lot more sense once we have select menus, textareas, and other form elements go inside of `FormGroup`s, which will be soon.

components/form_control/form_control.jsx
```diff
@@ -0,0 +1,88 @@
+const React = require('react');
+const classes = require('classnames');
+
+const responsiveClass = require('../_utilities/responsive_class.js');
+const size = require('../_utilities/size.js');
+
+function TextControl(props) {
+  return <input type="text" role="textbox" {...props} />;
```
sumul: Apparently, we're not supposed to apply `role="textbox"` to text inputs or textareas, since assistive devices already know that those are textboxes. I guess that role is for telling an assistive device that an unusual element is being used as a textbox. I'm also not even seeing that role in the latest lists of aria attributes. 

> Authors should avoid using this role in <input type="text"> or <textarea> elements. 

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_textbox_role)

components/form_control/form_control.jsx
```diff
@@ -0,0 +1,88 @@
...
+}
+
+function PasswordControl(props) {
```
sumul: So if we get rid of `role="textbox"` above, then the only difference between these various types is the `type` attribute. Here are all the HTML5 text types:

```
color
date
datetime
datetime-local
email
month
number
password
search
tel
text
time
url
week
```

Could we simplify the code to check whether `props.type` is one of those and call it a day?

components/form_control/form_control.jsx
```diff
@@ -0,0 +1,88 @@
...
+      onKeyUp={props.onKeyUp}
+      onPaste={props.onPaste}
+      pattern={props.pattern}
+      placeholder={props.placeholder}
+      readOnly={props.readOnly}
+      required={props.required}
+      size={props.size}
+      maxLength={props.maxLength}
+      value={props.value}
+      {...ariaProps(props)}
```
sumul: This spread is nice and clean. Why not do something similar with the list of props above?

components/form_control/form_control.jsx
```diff
@@ -0,0 +1,88 @@
...
+  const Control = getControlByType(props.type);
+
+  return (
+    <Control
+      autoComplete={props.autoComplete}
+      autoFocus={props.autoFocus}
+      className={classes(
+        props.className,
+        'FormControl',
+        'FormGroup__control',
```
sumul: Would it be possible not to embed the `FormGroup__control` class here? IMO this component should be self contained and usable without the `FormGroup` component. If you're implementing it within a `FormGroup` and have to apply the class manually, that would be fine (if not expected).

components/form_control/form_control.jsx
```diff
@@ -0,0 +1,88 @@
...
+        responsiveClass('FormControl--small', props, size.isSmall),
+        responsiveClass('FormControl--large', props, size.isLarge),
```
sumul: I'm wondering whether responsive utility classes should just be applied as regular classes, the way we do for general CSS responsive utilities. I see there's a discussion about this going on below. Will chime in there too.

components/form_group/README.md
```diff
@@ -0,0 +1,75 @@
+Form group component
```
sumul: I'd either omit this or flesh it out with more narrative. :-)

components/form_group/form_group.jsx
```diff
@@ -0,0 +1,62 @@
...
+const responsiveClass = require('../_utilities/responsive_class.js');
+const size = require('../_utilities/size.js');
+
+const FormControl = require('../form_control/form_control.jsx');
+
+class FormGroup extends React.Component {
+
+  renderChildren() {
+    // Preview environment won't have a child prop, so we have to supply a default one here:
+    if (this.props.__preview) {
```
sumul: I'm not sure this hack is necessary, and I'd like to avoid putting Fractal-specific stuff in our components if possible. Since your config for this component is a full JS file, I believe you can provide a `children` prop in the default context, which will get mapped to `this.props.children`. I don't think you can make the config file .jsx, so you might have to use `React.createElement` syntax, but that's not a huge deal.

guides/10-component-conventions.md
sumul: I've been thinking about this and feel that it would be pretty nice to unify the way our sizing modifiers and responsive utilities work: they are all just CSS classes. Something like:

```jsx
<FormGroup classNames="FormGroup--large-md" />...</FormGroup>
<h1 class="h3 h2-sm h1-md">...<h1>
```

Let's discuss!

components/form_group/_form_group.scss
```diff
@@ -0,0 +1,63 @@
...
+  @include form-group-sizer($size: 'small')
+}
+
+.FormGroup--large {
+  @include form-group-sizer($size: 'large')
+}
+
+@if $enable-type-utilities {
```
sumul: Do you think these should be lumped in with type utilities? We could easily create a new toggle for form utilities.

components/form_control/form_control.jsx
NickDunkman: Yea agree with this ... I think if you want to add one in a FormGroup you'd do this:

```jsx
<FormGroup>
  <TextInput className="FormGroup__text-input" />
</FormGroup>
```

sumul: I was thinking we'd still use `.FormGroup__control` to designate whatever is the control element, so either of these would work:

```jsx
<FormGroup>
  <TextInput className="FormGroup__control" />
</FormGroup>
```

or

```jsx
<FormGroup>
  <SelectMenu className="FormGroup__control" />
</FormGroup>
```

In this case, the `TextInput` and `SelectMenu` components would have their own styles and behaviors that have no knowledge of `FormGroup`. The `.FormGroup__control` class would allow the `FormGroup` to specify component-level positioning and maybe even modify or hook into behavior.

I'm still murky on error classes and stuff though. If we don't want the atomic elements (text input and select) to know about `FormGroup` (which I'm hoping is feasible), then error states and stuff that affect both the group and the control can be tricky if we maintain BEM purity. We might need to have some whitelisted generic state-specifying classes, like `.has-error` that can go on a `TextInput` or on a `FormGroup`, so you'd end up with some flexibility:

```scss
.text-input {
  // ...
  
  &.has-error,
  .has-error & {
    border-color: red;
  }
}
```

components/form_control/form_control.jsx
gelsto: I would like to do a spread here, especially since it's starting to feel like I'm re-implementing HTML (which I really, really do not want to do), but I'm wary about doing a `{...props}` spread since a recent(ish) React change throws warnings when you spread non-HTML/event handing props onto HTML elements. Being super explicit about the props allowed here means that you can do a `{...props}` spread upstream and not have any warnings, but I can see the argument that we should just expect the consumer to be in charge of filtering props before they get passed to the component.
In pare_web we have an `allowableHTMLProps` utility that filters out potentially problematic props .  We could implement a similar solution here, but (1) that was written as a reaction to a bunch of our code suddenly causing warnings, and I feel like there should be a more elegant solution, and (2) I wonder if in the future React will be even more specific wrt what props are allowed on what type of element (meaning a single utility won't work).

components/form_group/form_group.jsx
gelsto: You're probably right. I tried that, but I was having some issues with getting fractal to cooperate with changing imports (I'm sometimes unclear about when you have to rebuild vs. re-run `make dev` vs. refresh the page vs. expect hmr to take care of it). I'll take another look at it.

guides/10-component-conventions.md
gelsto: So the decision here is to just pass the full modifier as the className? Seems like that solution is consistent with the other comments below. I suppose if we're always naming the BEM base class after the component we're not doing anyone a huge favor by tacking on that suffix inside the component.

components/form_group/_form_group.scss
gelsto: Just to confirm: We should use medium size by default (which is already the case), but should probably add `.FormGroup--default-*` modifiers we can use if we need them.

components/form_group/form_group.jsx
gelsto: Turns out this doesn't work for a few reasons: In the server-side render, the context is copied using `_.cloneDeep`, which replaces functions with empty strings ([here](https://github.com/frctl/fractal/blob/master/src/core/resolver.js#L76)). If you somehow skip the server side render, the `Shuttle` component stringifies the context and removes the  `child` key all together. If you use `renderToString` or `renderToStaticMarkup` in the context, the html doesn't get escaped. I would rather do a check for the `__preview` prop than use `__dangerouslySetInnerHTML`. It would be really nice to be able to pass functions in context, though. I'm not sure how feasible that would be without making a change to the fractal adapter.

gelsto: We can discuss, but I think we should leave is as-is for this PR.

components/form_group/_form_group.scss
gelsto: dc28bd1

guides/10-component-conventions.md
gelsto: 489fb15

components/form_control/form_control.jsx
gelsto: 5cb949d

components/form_control/form_control.jsx
gelsto: I came up with this: e2b7d87. It requires us to be a little saavy with how we write the input sizing/error state mixins...

components/form_group/README.md
gelsto: merp derp 40a421d

components/form_group/_form_group.scss
gelsto: I removed the `type` toggle here: d576828. Do you think the form utilities will add enough CSS to justify adding a toggle for it?

components/text_input/text_input.jsx
```diff
@@ -0,0 +1,77 @@
...
+    if (key.match('^aria-')) {
+      return Object.assign(ariaAttrs, { [key]: props[key] });
+    }
+    return ariaAttrs;
+  }, {});
+}
+
+function TextInput(props) {
+  return (
+    <input
```
sumul: If you'll bear with me, I'd like to dwell on the way this component is implemented for a bit longer. I'm probably just not understanding some React things, but just want to make sure, since I'm expecting there will be other components in Phenotypes that lightly wrap a standard HTML element that we want to be able to pass any valid HTML attributes to, and this seems to set a fairly complex precedent for a component that doesn't do all that much. 

Is this the general workflow for such a component?

1. Do Reacty things with props that have nothing to do with the rendered HTML
2. Process and/or validate certain props before they're applied as HTML attributes (e.g. `type`, `className`)
3. Collect any other props that are valid HTML attributes and apply them to the HTML element

If that's roughly true, I have some things that I'd love to discuss. Namely: 

* What should happen when you pass a prop that doesn't apply to (1), (2), or (3) above? Does this implementation silently discard it?
* What's maintenance like as the HTML spec evolves?
* Should the onus of filtering arbitrary valid attributes fall on each component, or could we set up some style guide pattern that shifts it upstream and reduces the need to check everything so explicitly (e.g. passing in a specifically named object of dumb HTML attributes that are automatically spread)?
* How accurate should the filtering be? If we keep the onus on the components, then would it be sufficient to keep a master list of valid HTML attributes in a utility and let the consumer worry about making sure the specific attributes they're passing can actually be applied to the element?

components/form_group/form_group.jsx
gelsto: That would work, but it seems confusing to have it as a default prop since you would never use it in practice (since the input wouldn't have any event handlers on it). IMO we shouldn't have any default props that always have to overwritten.

sumul: Yeah, good point. Okay, let's do it this way for now. :-)

It would be cool to figure out how to pass functions in the context. I'll add that to the backlog.

components/text_input/README.md
```diff
@@ -41,18 +23,7 @@ url
...
-onChange
-onCopy
-onCut
-onFocus
-onInput
-onKeyDown
-onKeyUp
-onPaste
-```
+Valid HTML attributes and event handlers for the input element may be safely passed as additional props. Passing a prop that is not a legal DOM attribute will cause React to trigger an [unknown prop warning](https://facebook.github.io/react/warnings/unknown-prop.html).
```
sumul: Reminder about camel casing would be helpful here.

components/form_group/_form_group.scss
```diff
@@ -0,0 +1,69 @@
...
+      font-size: $font-size-4;
+    }
+
+    .FormGroup__hint {
+      font-size: $font-size-2;
+    }
+  }
+
+  .FormGroup__control {
+    @include text-input-sizer($size: $size);
```
sumul: On second thought, I don't think we should embed this responsive utility here. Since each of the `.FormGroup__control` components will already need to have its own responsive utilities for this to work, this is going to create potentially a TON of redundant CSS (# breakpoints X # 3 sizes X # of control components). The benefit of not having to remember to responsively size the embedded component through classes isn't worth the tradeoff IMO.

components/text_input/README.md
gelsto: Good point - 	623ff3d

components/text_input/README.md
```diff
@@ -0,0 +1,36 @@
...
+password
+search
+tel
+text
+time
+url
+week
+```
+
+Valid HTML attributes and event handlers for the input element may be safely passed as additional props. Passing a prop that is not a legal DOM attribute will cause React to trigger an [unknown prop warning](https://facebook.github.io/react/warnings/unknown-prop.html).
```
sumul: Reminder about camel case here too?

components/text_input/README.md
```diff
@@ -0,0 +1,38 @@
...
+tel
+text
+time
+url
+week
+```
+
+Valid HTML attributes* and event handlers for the input element may be safely passed as additional props. Passing a prop that is not a legal DOM attribute will cause React to trigger an [unknown prop warning](https://facebook.github.io/react/warnings/unknown-prop.html).
+
+**Note that there are [several attributes that work differently between React and HTML](https://facebook.github.io/react/docs/dom-elements.html). In these cases, attributes should be passed the React-y way.*
```
sumul: Little markdown typo? Looks like the rest of the README is bold now.

components/text_input/text_input.jsx
```diff
@@ -0,0 +1,46 @@
+const React = require('react');
+const classes = require('classnames');
+
+const INPUT_TYPES = {
```
sumul: Just curious: is it better for this to be an object of keys and values instead of an array of valid strings to check against?

gelsto: Object lookup is constant time so this is should be more efficient than looping through a list. Doesn't really matter in this case.

components/form_group/_form_group.scss
gelsto: Can we put it in another toggle? Seems like it would be really nice to be able to style the control with the group...

PR #23, opened 2017-06-20
Feature/button

gelsto:
@sumul - JS/BEM modifiers for button component
I broke up the logic for rendering achor-style buttons vs. regular buttons into separate files, but the only code available under `View` on preview is the `button/button.jsx` file. Is there a way to modify this? I feel like it would be really noisy in one file.

### DIFF
--- a/components/button/README.md
+++ b/components/button/README.md
@@ -1,7 +1,19 @@
 ### Props
 
+If a button has an `href` prop, the component renders an achor tag with `role=button`. Otherwise, the button is rendered as a true `button` HTML element.
+
 | name | type | default | description |
 | ---- | ---- | ------- | ----------- |
+| `className` | string | | Class name for the root element. This should include any BEM modifier classes you wish to use.
+| `disabled` | boolean | | If `true`, renders the button in a disabled state and adds an `aria-disabled` attribute for assistive technologies. For true buttons, sets the `disabled` attribute on the element. For anchor-style buttons, adds `tabIndex=-1` to prevent user interaction and disables click handling/pointer events.
+| `href` | string | | Determines if the component renders an anchor-style button or a true button. If the `href` prop is `'#'`, `event.preventDefault()` is automatically called on click events.
+| `onClick` | function | | Handler for click events. Called with the [SyntheticEvent](https://facebook.github.io/react/docs/events.html) as the only argument.
+| `type` | one of: `'button'`, `'submit'`, `'reset'` | `'button'` | Passed `type` attribute for the `button` element. Ignored by anchor-style buttons.
+
+
+Valid HTML attributes and additional event handlers for the `button` or `a` element tags may be safely passed as extra props. Passing a prop that is not a legal DOM attribute will cause React to trigger an [unknown prop warning](https://facebook.github.io/react/warnings/unknown-prop.html). Additionally, you should take care that the props you're passing make sense for the element type being rendered (for example, you should only be passing a `target` or `rel` prop to an anchor-style button).
+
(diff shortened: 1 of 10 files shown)


components/button/README.md
```diff
@@ -0,0 +1,44 @@
...
+Valid HTML attributes and additional event handlers for the `button` or `a` element tags may be safely passed as extra props. Passing a prop that is not a legal DOM attribute will cause React to trigger an [unknown prop warning](https://facebook.github.io/react/warnings/unknown-prop.html). Additionally, you should take care that the props you're passing make sense for the element type being rendered (for example, you should only be passing a `target` or `rel` prop to an anchor-style button).
+
+*Note that there are [several attributes that work differently between React and HTML](https://facebook.github.io/react/docs/dom-elements.html). In these cases, attributes should be passed the React-y way.*
+
+
+### Modifiers
+
+| class | default | responsive | notes |
+| ----- | ------- | ---------- | ----- |
+| `Button--default` | true | true | |
```
sumul: `--medium`, no?

components/button/_button.scss
```diff
@@ -0,0 +1,48 @@
+// scss-lint:disable SelectorFormat
+@mixin button-sizer($size: 'medium') {
+  &.Button {
```
sumul: When this mixin is used below, it gets compiled as e.g. `.Button--small.Button`. Is that intentional?

components/button/_button.scss
```diff
@@ -0,0 +1,48 @@
...
+
+.Button--small {
+  @include button-sizer($size: 'small');
+}
+
+.Button--large {
+  @include button-sizer($size: 'large');
+}
+
+@each $breakpoint in map-keys($grid-breakpoints) {
```
sumul: Let's put this under the form utilities toggle.

components/button/button.config.js
```diff
@@ -0,0 +1,90 @@
+module.exports = {
+  name: 'button',
+  label: 'Button',
+  status: 'draft',
+  collated: true,
+  context: {
+    children: 'Button',
```
sumul: It would be nice to have the text in the preview for each button variant describe itself (e.g. "Disabled button", "Disabled primary button", etc.).

components/button/button.jsx
```diff
@@ -0,0 +1,22 @@
...
+const TrueButton = require('./true_button.jsx');
+
+function Button(props) {
+  const Component = props.href ? AnchorButton : TrueButton;
+  const isDisabled = !!props.disabled;
+
+  return (
+    <Component
+      {...props}
+      aria-disabled={isDisabled}
```
sumul: I noticed that `aria-disabled` always shows up as an attribute in the rendered HTML, even if it's `false`. Do we need to show it when it's `false`?

components/button/README.md
gelsto: Blerg.

components/button/button.jsx
gelsto: Nope: 2a71dba

components/button/README.md
gelsto: 🔔 0e37fc2

components/button/_button.scss
gelsto: That's leftover from the pattern I was using for the responsive text inputs. 5738b6e

components/button/_button.scss
gelsto: f7c4a11

components/button/button.config.js
gelsto: c84aaa7

PR #30, opened 2017-07-06
TextInput and FormGroup styles

sumul:
`TextInput` styles:
screen shot 2017-07-06 at 6 23 33 pm

`FormGroup` styles:
screen shot 2017-07-06 at 6 23 50 pm

Temporarily using a red ✘ in place of the SVG alert icon for the error state.

### DIFF
--- a/components/form_group/_form_group.scss
+++ b/components/form_group/_form_group.scss
@@ -1,39 +1,54 @@
-// scss-lint:disable SelectorFormat
+// scss-lint:disable SelectorFormat ElsePlacement
 @mixin form-group-sizer($size: 'medium') {
   @if $size == 'small' {
-    .FormGroup__label,
-    .FormGroup__error,
-    .FormGroup__hint {
-      font-size: $font-size-1;
+    .FormGroup__label {
+      @include text-1($font-weight-bold);
+      margin-bottom: $spacer-1;
+    }
+
+    .FormGroup__error {
+      @include text-1;
     }
   }
 
-  @if $size == 'medium' {
-    .FormGroup__label,
+  @else if $size == 'medium' {
+    .FormGroup__label {
+      @include text-2($font-weight-bold);
+      margin-bottom: $spacer-2;
+    }
+
     .FormGroup__error {
-      font-size: $font-size-2;
+      @include text-2;
+      color: $brand-danger;
+      margin-top: $spacer-2;
+
+      &::before {
+        content: '✘ '; // TODO: replace with SVG icon
+        display: inline;
+      }
     }
 
     .FormGroup__hint {
-      font-size: $font-size-1;
+      @include text-1;
+      color: $text-color-secondary;
+      margin-top: $spacer-2;
     }
   }
 
-  @if $size == 'large' {
-    .FormGroup__label,
+  @else if $size == 'large' {
+    .FormGroup__label {
+      @include text-4($font-weight-bold);
+    }
+
     .FormGroup__error {
-      font-size: $font-size-4;
+      @include text-3;
     }
 
     .FormGroup__hint {
-      font-size: $font-size-2;
+      @include text-2;
     }
   }
 }
-
-.FormGroup__error { color: $red; }
-
-.FormGroup__label { font-weight: $font-weight-bold; }
 
 //
 // Sizers
--- a/components/form_group/form_group.config.js
+++ b/components/form_group/form_group.config.js
@@ -39,10 +39,28 @@ module.exports = {
       error: 'Error message',
     },
   }, {
+    name: 'error small',
+    label: 'Error (small)',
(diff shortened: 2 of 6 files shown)


components/form_group/_form_group.scss
```diff
@@ -1,40 +1,55 @@
...
 @mixin form-group-sizer($size: 'medium') {
   @if $size == 'small' {
-    .FormGroup__label,
-    .FormGroup__error,
-    .FormGroup__hint {
-      font-size: $font-size-1;
+    .FormGroup__label {
+      @include text-1($font-weight-bold);
+      margin-bottom: $spacer-1;
+    }
```
iqnivek: IMO it's a little confusing to have a label option on the FormGroup, because if there are multiple inputs in props.children, providing a label is now sort of wonky/invalid, because that label should only apply to the first input, but other inputs would need to provide their own labels in props.children.

If instead that label is supposed to be a header for all the child inputs, then I think a different tag should be used, because [labels shouldn't be used as a header](https://stackoverflow.com/questions/33214093/is-it-good-practice-to-use-label-for-non-input-non-interactive-elements). But there are many ways you could want a grouping header, so I think that tag should also be specified outside of a FormGroup for flexibility.

So my suggestion is to remove the `label` and `controlId` props from FormGroup, and specify any labels / headers as part of props.children instead.

NickDunkman: @gelsto ^

sumul: @iqnivek I chatted at length about this with @gelsto when the component's behavior was implemented. @gelsto decided on a pattern where the component would provide the `label` and connect it to a single input if you provide the `controlId` of the input. If you omit the `controlId` then you are free to supply your own labels and connections in the children. In the latter case, you end up with a `<label>` HTML element that's not connected to any inputs, but I'm okay with the semantics of this. 

I think I'm describing this correctly. It's documented thoroughly in the README for this component with examples.

iqnivek: I still think it's confusing to have props that should only be used with a certain child structure (i.e. a single input).. I think it'd be fine if the component could only wrap a singular input, but the component is named a Form**Group**. 

That being said, I can live with it if you guys feel strongly about it, and it's definitely not a blocker for this PR.

gelsto: @iqnivek and I had a brief discussion about this yesterday. If the weirdness stems from a `label` that might not have a `for` attribute, we could just render a different html element if we don't have a `controlId` prop.

Maybe a solution is to have something like `FormInput` that requires a singular input and renders the wrapper/label/hint/error (what `FormGroup` is now, but requires a `controlID`), and `FormGroup` that doesn't have specific child input requirements and renders only the wrapper/hint/error.  Both would have to take the same modifiers. So:
```es6
function FormInput(props) {
  const { label, controlId, children, ...groupProps } = props;
  return (
    <FormGroup {...groupProps}>
      <label htmlFor={controlId} className="FormGroup__label">{label}</label>
      { children }
    </FormGroup>
  );
}

class FormGroup extends React.Component {
  // ...
  render() {
    return (
      <div
        className={classes(
          this.props.className,
          'FormGroup',
          { 'FormGroup--has-error': !!this.props.error }
        )}
      >
        { this.props.children }
        { this.renderError() }
        { this.renderHint() }
      </div>
    );
  }
}
```

gelsto: I think this is out of scope for this PR, though.

components/form_group/_form_group.scss
```diff
@@ -1,40 +1,55 @@
...
   }
 }
 
-.FormGroup__error { color: $red; }
-
-.FormGroup__label { font-weight: $font-weight-bold; }
-
 //
 // Sizers
 //
```
iqnivek: this wasn't introduced in this PR, but these 3 lines that're below in this file shouldn't be needed, should they?

```
@include form-group-sizer;

.FormGroup--small { @include form-group-sizer($size: 'small'); }
 
.FormGroup--large { @include form-group-sizer($size: 'large'); }
```

I think those classes are gonna be redeclared in the `@each $breakpoint in map-keys($grid-breakpoints) {` loop because the `breakpoint-suffix` mixin automatically makes the suffix blank ("") for the lowest -xs breakpoint.

sumul: I think they need to be there in case the responsive classes are toggled off.

sumul: It's a good point though. I'll update this and similar things to only include these redundant sizers if the responsive feature is toggled off.

sumul: @iqnivek https://github.com/aminohealth/phenotypes/pull/31/commits/89a1ac1069ae84741220195328c0286f533d078d

PR #31, opened 2017-07-07
Checkbox component

sumul:
screen shot 2017-07-07 at 5 29 15 pm

@NickDunkman @gelsto @iqnivek Here's a checkbox component. I mostly focused on getting the styles right. The React component is based on similar patterns as the TextInput and FormGroup components, but we might want to change things or flesh it out depending on how we want to use this. 

If we need to make significant changes to the React component, I'd prefer one of y'all take over development of this component and address those in a follow-up PR.

@iqnivek This PR also addresses the feedback you posted here: https://github.com/aminohealth/phenotypes/pull/30#discussion_r126238071.


NickDunkman:
🚀

PR #38, opened 2017-07-13
Make header margins 0.5rem?

iqnivek:
I noticed that the margin-bottom was changed from 1rem to 0.5rem in the reboot. I don't know all the potential use cases, but at least for the new marketing copy, using 0.5rem would allow not having to customize the headers. Don't feel super strongly though if you expect 1rem to be more applicable in the future. Alternatively, if the headers on the 3-column section can just go with 1rem instead of 0.5, that would work too.

Comparison of 1rem to 0.5rem in our current marketing stuff:

1rem 0 5rem

### DIFF
--- a/styles/_reboot.scss
+++ b/styles/_reboot.scss
@@ -104,7 +104,7 @@ hr {
 // margin for easier control within type scales as it avoids margin collapsing.
 h1, h2, h3, h4, h5, h6 {
   margin-top: 0;
-  margin-bottom: 1rem;
+  margin-bottom: 0.5rem;
 }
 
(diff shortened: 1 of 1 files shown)


sumul:
@iqnivek I'm not sure either default gets us everything we need to match the design spec. In the design, at `xs` size, the big header has 16px margin and the small headers have 0. At larger sizes, it's 16px and 5px. 0.5rem is 8px, which is not a standard spacer value, so I'd hesitate to make it the default.

iqnivek:
okay, that makes sense - thanks sumul!

PR #40, opened 2017-07-17
Form element styling improvements and fixes for mobile

sumul:
@iqnivek @NickDunkman There are a lot of visual issues with `Button` and `TextInput` on mobile webkit. This PR improves styling and fixes some bugs.

### DIFF
--- a/fractal_assets/js/theme.js
+++ b/fractal_assets/js/theme.js
@@ -1,0 +1,1 @@
+document.addEventListener("touchstart", function() {},false);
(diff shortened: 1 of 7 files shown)


fractal_assets/js/theme.js
```diff
@@ -0,0 +1 @@
+document.addEventListener("touchstart", function() {},false);
```
iqnivek: I know this is in the button docs, but would suggest adding a comment here too mentioning that this is for signaling the browser to apply touch styles.

NickDunkman: Gonna merge this and then add this comment in my own upcoming PR

iqnivek: 👍

PR #41, opened 2017-07-18
Reorganize components in Fractal

NickDunkman:
cc @sumul 

See screenshot below for the new structure!

- Tweaked our `fractal_react` react component generator to allow for collections. It's recursive, so we can have collections inside of collections inside of collections etc.
- The utility css is now co-located with the demo components, like we do with our regular component css.
- Fixed a bunch of linting errors and converted some tabs to spaces
- The `grid` css is now considered a utility
- Renamed our "text" utility grouping to "typography". That's what we call it in the "guides" section, and it's also what Bootstrap calls [their version](http://getbootstrap.com/css/#type) of that documentation. (none of the css classes themselves changed, tho ... just the fractal component name).

There's one thing that's not backwards compatible -- the "component" react file paths have slightly changed, so we'll need to update pare_web import paths when we upgrade. But that's pretty trivial.

screen shot 2017-07-18 at 16 15 56


library/utilities/display/display.jsx
```diff
@@ -1,23 +1,23 @@
-import React from 'react';
```
iqnivek: if there are no other considerations, I think it'd be better to move to using ES6 syntax for phenotypes imports. I realize we're probably more consistently using `require` right now, so if we do it probably makes sense to do in a separate pass.

library/utilities/typography/typography.jsx
```diff
@@ -127,4 +132,4 @@ function TextUtilities(props) {
   );
 }
 
-module.exports = TextUtilities;
```
iqnivek: in the same vein as the ES6 imports comment earlier, I think we should move to ES6-style exports as well if possible. Not a blocker, just raising that as a potential thing to look at.

library/utilities/display/display.jsx
NickDunkman: Ok - agree that sounds better. I made that change b/c it fixed linting errors. Maybe someone can just fix the linting situation + convert all these at once in the future.

NickDunkman: Actually, gonna go ahead and convert, since you have to do that in order to enable the es6 exporting. We can just live with linting errors for now 😀

NickDunkman: Nah - this looks like a lot of work. Let's do it down the road.

NickDunkman: https://www.pivotaltracker.com/story/show/149127403

PR #42, opened 2017-07-19
Add more utilities

NickDunkman:
Brings some utilities over from pare_web that I think will be commonly used enough to be here.

- sizing utilities, such as `.width5`, `.height3-sm` and also some 100% utils like `.width-full` and `.height-full`.
- negative margin classes (e.g. `mtn3`)
- `.text-underline`
- `.flex-none`, for specifying that an item in a flexbox should not shrink or grow. Basscss has this class.

cc @sumul 

sizing

### DIFF
--- a/fractal_assets/css/preview.css
+++ b/fractal_assets/css/preview.css
@@ -13,3 +13,7 @@
   background-color: #2dcfa1;
   border: 1px solid #007868;
 }
+
+.sizing-example {
+  background-color: #2dcfa1;
+}
--- a/fractal_react/__components.js
+++ b/fractal_react/__components.js
@@ -11,6 +11,7 @@ module.exports = {
   "display": require("../library/utilities/display/display.jsx"),
   "flexbox": require("../library/utilities/flexbox/flexbox.jsx"),
   "grid": require("../library/utilities/grid/grid.jsx"),
+  "sizing": require("../library/utilities/sizing/sizing.jsx"),
   "spacing": require("../library/utilities/spacing/spacing.jsx"),
   "typography": require("../library/utilities/typography/typography.jsx"),
 };
--- a/library/components/button/_button.scss
+++ b/library/components/button/_button.scss
@@ -187,7 +187,7 @@
 // Sizers and responsive utilities
 // -------------------------------
 
-@if $enable-form-utilities {
+@if $enable-responsive-form-components {
   @each $breakpoint in map-keys($grid-breakpoints) {
     @include media-breakpoint-up($breakpoint) {
       $suffix: breakpoint-suffix($breakpoint, $grid-breakpoints);
--- a/library/components/checkbox/_checkbox.scss
+++ b/library/components/checkbox/_checkbox.scss
@@ -123,7 +123,7 @@
   user-select: none;
 }
 
-@if $enable-form-utilities {
+@if $enable-responsive-form-components {
   @each $breakpoint in map-keys($grid-breakpoints) {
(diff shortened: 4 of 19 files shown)


sumul:
@NickDunkman Cool! I'd like to be an explicit reviewer on this. There are several utilities I wanted to port over from Bootstrap too, so I want to make sure we're not duplicating things and check for consistency with their patterns. I can review tomorrow morning first thing.

sumul:
@NickDunkman After a quick read-through, this looks really good. Two thoughts:

1. Do you think we shouldn't have any percentage-based sizing utilities? Is that best left to the grid? You mentioned that the grid is preferred for setting widths, but I just want to make sure I'm understanding the intent of the utilities correctly. Seems like it could be handy to have a few percentages (25%, 33%, 50%, 67%, 75%, 100%), but I'm happy to keep this focused.
2. The `width3` syntax looks weird to me. It would feel more consistent with other classes if it were like `text-1` (`width-3`) or `m3` (`w3`). Tricksy.

NickDunkman:
@sumul I'm not completely opposed to adding other percentage widths (33%, 50%, etc), but I think should see if we actually end up ever using those, first. May end up just using the grid exclusively for percentage-based widths.

My comment about preferring the grid vs. fixed width classes is related to the fixed pixel-based width classes (width0 - width9). 

Using `w3` and `h3` conflicts with the <h1> classes like `.h1`. But I like how our spacing classes (like `m3` and `pt4-sm`) don't have that extra dash in there. Wish we would have done that with the text classes as well.

NickDunkman:
@sumul better? https://github.com/aminohealth/phenotypes/pull/42/commits/02d4697cb510517a3733577b657a604535acb3ee

Might have been a little heavy-handed as it was.

iqnivek:
Yeah, I agree that most of the time sub-100% widths can likely be handled by grid (bootstrap does do 25%, 50%, and 75%, but I think we can hold off until necessary too).

I do kinda agree with Sumul that we should either use the `.width-1` naming convention or `.w1`, but not the mixed format `.width1`. I'm leaning toward `.width-1` for explicitness - while the spacers are nice and compact, I think we should be careful not to go too far with compatifying everything, and width- and height- are reasonably short without writing them out.

iqnivek:
Otherwise looks good to me! I'll let Sumul handle final approval.

NickDunkman:
@sumul @iqnivek added dashes here: https://github.com/aminohealth/phenotypes/pull/42/commits/f0ca19588377784e1b1cd332c2a529aace1f8e7e

Having trouble deciding on the right naming schema for the percentage classes, so think we should pass for now. Different variations considered:

```
.width-full
.width-half
.width-fourth
.width-third
.width-three-fourths
.width-two-thirds
```

Those are nice b/c you can just describe an infinite decimal percentage like 33.33333% as just "third". But they're annoying when it's anything other than `1/x`.

```
.width-1/2
.width-1/3
.width-1/4
.width-3/4
```

From a pure css standpoint, I like these best, but the slashes could cause some headaches with js/css tooling.

```
.width-50p
.width-25p
.width-75p
.width-33p
```

The 33p thing is annoying here.

Still adding `.width-full` for 100% for now ... we need those and I think I like having that class even if we decide to also have something like `.width-100p`.

NickDunkman:
@sumul Added a demo, & also added the `.flex-1` class, which causes a flexbox child to expand to fill the horizontal space of the flexbox.

screen shot 2017-07-20 at 09 48 49

PR #45, opened 2017-07-20
Fix scss compilation in other projects

NickDunkman:
@iqnivek I changed these import paths when I did the component reorg, without realizing that it would break things when another project (like pare_web) tries to compile the css.


iqnivek:
👍

PR #49, opened 2017-10-25
Add radio components

NickDunkman:
These are similar to checkboxes--they have the same sizes, use the hidden `<input>` trick for accessibilities, and there are no animations.

radios


sumul:
radio___phenotypes

sumul:
Something weird is going on with keyboard navigation. On the checkboxes, you can tab through all the enabled elements in order. With the radios, it seems to tab to the first one and then skips to the button at the bottom.

library/components/radio/_radio.jsx
```diff
@@ -0,0 +1,186 @@
...
+      <h6>States &amp; modifiers</h6>
+      <div>
+        <Radio className="mr4">Unchecked</Radio>
+        <Radio defaultChecked>Checked</Radio>
+      </div>
+      <div>
+        <Radio className="mr4" disabled>Disabled</Radio>
+        <Radio disabled defaultChecked>Disabled and checked</Radio>
+      </div>
+      <Radio className="d-block mt5 Radio--small" defaultChecked><code>Radio--small</code></Radio>
```
sumul: Setting these in `<code>` makes the preview look weird IMO. Might be better to match the other previews (e.g. checkbox, which just shows some sample descriptive text instead of class names). That makes it easier to compare to the design too.

library/components/radio/radio_group.jsx
```diff
@@ -0,0 +1,42 @@
...
+    const newValue = clearing ? null : valueOfRadioClicked;
+
+    if (newValue !== this.state.value) {
+      this.setState({ value: newValue });
+      this.props.onChange && this.props.onChange(newValue);
+    }
+  }
+
+  render() {
+    return this.props.render({
```
gelsto: Since these options are used identically for every radio, it might make sense to pass a function here instead:
I did some playing around, and you can do something like this:
```jsx
render() {
  const GroupRadio = (radioProps) => {
    const checked = radioProps.value === this.state.value;
    return (
      <Radio
        {...radioProps}
        checked={checked}
        disabled={this.props.disabled || radioProps.disabled}
        onChange={() => {
          if (checked && !this.props.clearable) {
            return;
          }
          const newValue = checked ? null : radioProps.value;
          this.setState({ value: newValue });
          this.props.onChange(newValue);
        }}
      />
    );
  };

  return this.props.render(GroupRadio);
}
```

Then the consumer uses it as:
```jsx
function BasicGroupExample() {
  return (
    <div>
      <p>
        The <code>RadioGroup</code> component makes it easy to have a single field with a selected
        value, using multiple radio buttons.
      </p>
      <RadioGroup
        value="two"
        render={(Radio) => (
          <div className="mt5">
            <Radio
              className="mr4"
              value="one"
            > One
            </Radio>
            <Radio
              className="mr4"
              value="two"
            > Two
            </Radio>
            <Radio
              value="three"
            > Three
            </Radio>
          </div>
        )}
      />
    </div>
  );
}
```

@iqnivek I'm wondering what you think of this ^? I haven't seen it before but it seems like passing a pre-configured component would make sense here....

library/components/radio/radio_group.jsx
```diff
@@ -0,0 +1,42 @@
...
+      disabled: this.state.disabled,
+      onChange: this.handleChange,
+    });
+  }
+}
+
+RadioGroup.defaultProps = {
+  value: null,
+  disabled: false,
+  clearable: false,
```
gelsto: Should `onChange` be a default prop here?

library/components/radio/_radio.scss
```diff
@@ -0,0 +1,139 @@
+@mixin radio-sizer($size: 'medium') {
+
+  // medium
+  $indicator-width: modular-scale-px(0);
+  $indicator-top: 1px;
+  $dot-width: 6px;
+  $dot-nudge: -50%;
+  $spacing: $spacer-2;
```
sumul: Let's bump each the spacer value down by one for each size. Since the radio is a circle, it visually appears farther away from the text than the straight edge of the square checkbox.

library/components/radio/_radio.jsx
```diff
@@ -0,0 +1,186 @@
...
+  return (
+    <div className="mt5">
+      <h6>Sophisticated labels</h6>
+      <p>
+        You can use html in the label if it needs to be more than just a string. Clicking anywhere
+        in the label will check the radio.
+      </p>
+      <Radio>
+        I have some footer text
+        <p className="text-1 text-color-secondary">click me</p>
```
sumul: This is cool. Does it work on checkboxes too?

NickDunkman:
@sumul The tab thing is because I'm not putting the `name` attribute on the radio buttons in the demo - although you do have the option of doing that, with the way I designed it.

I'll add those so that you can see that behavior in the demo.

library/components/radio/_radio.scss
```diff
@@ -0,0 +1,139 @@
...
+  } @else if $size == 'large' {
+    $indicator-width: modular-scale-px(2);
+    $indicator-top: 2px;
+    $dot-width: 8px;
+    $dot-nudge: -55%;
+    $spacing: $spacer-3;
+    $text-size: 4;
+  }
+
+  @include text-size($text-size);
```
sumul: Does organizing the code this way get us something beyond reducing redundancy. I think I prefer the pattern that we used for the checkbox component. It's more verbose and repetitive, but I find its explicitness a lot easier to understand. And it doesn't require this little hack (which I'd love to avoid needing if possible).

NickDunkman:
@sumul this should fix tabbing for you: https://github.com/aminohealth/phenotypes/pull/49/commits/caae617708c414c759300ba5969caa8a7c9b8e33

library/components/radio/radio_group.jsx
NickDunkman: That is fancy af.

NickDunkman: I like how that removes the need to import the regular `Radio` component yourself when you're creating a group.

library/components/radio/_radio.scss
NickDunkman: Cool - what do you think about the case where we show checkboxes & radios adjacent to eachother vertically. Would it look weird for the labels to not line up?

library/components/radio/_radio.scss
NickDunkman: That works for me. Curious why you don't like the `text-size` mixin though - could you explain why? Doesn't seem harmful to me.

library/components/radio/radio_group.jsx
```diff
@@ -0,0 +1,42 @@
+const React = require('react');
+
+class RadioGroup extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+    this.state = {
```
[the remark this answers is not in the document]
iqnivek: definitely agree that RadioGroup should be a controlled component with a `value` prop passed in

library/components/radio/_radio.jsx
NickDunkman: Our checkboxes use a `label` prop instead of the `children` prop - I think technically you could pass html down through the label prop too. But I prefer this way, it reads nicer.

Should we change checkboxes to match & then make a new major version?

@gelsto pointed out that we have the `FormGroup` component for having "hints" beneath fields - but I like that when you put the hint inside the label, clicking it checks the radio.

library/components/radio/radio_group.jsx
iqnivek: hmm, interesting approach! My hesitations are that it seems a little more magical and also is definitely a novel pattern, so but I can see it being easier to work with in most cases, so I'm down to try it.

library/components/radio/radio_group.jsx
NickDunkman: It's optional, thinking it's fine for it to just be `undefined`

library/components/radio/radio_group.jsx
NickDunkman: 🔥 https://github.com/aminohealth/phenotypes/pull/49/commits/b12b1913439b89df6524350fc44827ec83bb9cb4

library/components/radio/radio_group.jsx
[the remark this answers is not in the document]
NickDunkman: Changed to be stateless in this: https://github.com/aminohealth/phenotypes/pull/49/commits/b12b1913439b89df6524350fc44827ec83bb9cb4

library/components/radio/_radio.jsx
NickDunkman: Good point - changed in here: https://github.com/aminohealth/phenotypes/pull/49/commits/b12b1913439b89df6524350fc44827ec83bb9cb4

library/components/radio/radio_group.jsx
```diff
@@ -0,0 +1,40 @@
...
+        disabled={groupProps.disabled || radioProps.disabled}
+        onChange={() => {
+          // Call the radio's own onChange if it has one.
+          radioProps.onChange && radioProps.onChange();
+
+          // The group's onChange callback is only called when the value actually changes.
+          if (groupProps.onChange) {
+            if (!checked || groupProps.clearable) {
+              const newValue = checked ? null : radioProps.value;
+              if (newValue !== groupProps.value) {
```
gelsto: Do we need this check?
Since `checked` already reflects `radioProps.value`, I don't think we can get to this statement if the new value is the same as the current one....

.eslintrc
```diff
@@ -13,6 +13,7 @@
     "func-names": "off",
     "import/no-extraneous-dependencies": "off",
     "import/no-unresolved": "off",
+    "jsx-a11y/label-has-for": "off",
```
gelsto: 👍

library/components/radio/_radio.scss
sumul: I don't think it's harmful, but it does encourage writing fancy code that's harder to read and maintain. We already have mixins like `@include text-1`, and the syntactic flexibility that `@include text-size(1)` gets you ideally wouldn't be required to style simple components. I'd be into it if it allowed us to do something clever at the framework level, but so far we haven't needed it there.

library/components/radio/_radio.scss
NickDunkman: changed here: https://github.com/aminohealth/phenotypes/pull/49/commits/c64ec9d9be2d639576a2ffe69e8ffb6635737042

library/components/radio/_radio.scss
NickDunkman: Changed here: https://github.com/aminohealth/phenotypes/pull/49/commits/c64ec9d9be2d639576a2ffe69e8ffb6635737042

Instead of one big mixin with conditionals, I think it's better as separate `radio-small`, `radio-medium`, and `radio-large` mixins. Then you can't accidentally call with the wrong size string, and we have smaller mixins.

library/components/radio/radio_group.jsx
NickDunkman: Ya ur right, removed: https://github.com/aminohealth/phenotypes/pull/49/commits/57b8d1994229865da70e6adac3570cca30cf92c8

NickDunkman:
@sumul @gelsto Think I've addressed everything!

gelsto:
This is looking great 👍 - I'll leave to @sumul to approve/merge

NickDunkman:
Good catch @sumul, that was a tricky one to fix (w/ @gelsto's help). Should be good now.

PR #50, opened 2017-10-26
Feature/switch

gelsto:
switch

### DIFF
--- a/library/components/switch/README.md
+++ b/library/components/switch/README.md
@@ -1,0 +1,28 @@
+The `Switch` component renders a specialized `<input type="checkbox">` that is useful for displaying an "on" or "off" selection.
+
+### Props
+
+| name | type | default | description |
+| ---- | ---- | ------- | ----------- |
+| `className` | string | | Class name for the root element.
+| `isOn` | boolean | false | If the switch is toggled on. If set to `True`, the underlying `input` element will have a `checked=true` attribute.
+
+Valid HTML attributes and event handlers for the input element (ex. `disabled`, `onChange`) may be safely passed as additional props. **These props will be applied to the `<input>` element, not the root element of the component.**
+
+Passing a prop that is not a legal DOM attribute will cause React to trigger an [unknown prop warning](https://facebook.github.io/react/warnings/unknown-prop.html).
+
+*Note that there are [several attributes that work differently between React and HTML](https://facebook.github.io/react/docs/dom-elements.html). In these cases, attributes should be passed the React-y way.*
+
+### Usage
+
+Javascript:
+```html
+<Switch isOn />
+```
+Rendered markup:
+```html
+<label class="Switch">
+  <input class="Switch__input" type="checkbox" checked />
+  <div className="Switch__indicator" />
+</label>
+```
--- a/library/components/switch/README.md
+++ b/library/components/switch/README.md
@@ -5,19 +5,13 @@ The `Switch` component renders a specialized `<input type="checkbox">` that is u
 | name | type | default | description |
 | ---- | ---- | ------- | ----------- |
 | `className` | string | | Class name for the root element.
-| `isOn` | boolean | false | If the switch is toggled on. If set to `True`, the underlying `input` element will have a `checked=true` attribute.
-
-Valid HTML attributes and event handlers for the input element (ex. `disabled`, `onChange`) may be safely passed as additional props. **These props will be applied to the `<input>` element, not the root element of the component.**
-
-Passing a prop that is not a legal DOM attribute will cause React to trigger an [unknown prop warning](https://facebook.github.io/react/warnings/unknown-prop.html).
-
-*Note that there are [several attributes that work differently between React and HTML](https://facebook.github.io/react/docs/dom-elements.html). In these cases, attributes should be passed the React-y way.*
+| HTML attributes | various | | e.g. `checked: true`, `disabled: true`, events, etc. are applied to the `input` |
 
 ### Usage
 
 Javascript:
 ```html
-<Switch isOn />
+<Switch checked />
 ```
 Rendered markup:
 ```html
--- a/library/components/switch/_switch.scss
+++ b/library/components/switch/_switch.scss
@@ -1,5 +1,3 @@
-// scss-lint:disable SelectorFormat ElsePlacement StringQuotes
-
 @mixin switch-sizer() {
   $width: modular-scale-px(7);
   $height: modular-scale-px(3);
@@ -34,7 +32,7 @@
     @include depth(7);
     background: $white;
     border-radius: 50%;
-    content: " ";
+    content: ' ';
     position: absolute;
     top: $border-width;
     transition: right .15s ease-out;
@@ -55,7 +53,7 @@
   }
 
   &:disabled ~ .Switch__indicator {
-    cursor: auto;
+    cursor: not-allowed;
     opacity: $alpha-600;
   }
 }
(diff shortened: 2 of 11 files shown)


library/components/switch/_switch.scss
```diff
@@ -0,0 +1,72 @@
...
+}
+
+.Switch__indicator {
+  background-color: $gray-300;
+  transition: background-color .2s linear;
+
+  &::after {
+    background: $white;
+    border-radius: 50%;
+    box-shadow: 0 5px 10px 0 rgba($black, $alpha-300), 0 1px 2px 0 rgba($black, $alpha-200), 0 0 10px 0 rgba($black, $alpha-200);
```
sumul: `@include depth(7);`
https://github.com/aminohealth/phenotypes/blob/dev/styles/mixins/_depth.scss

library/components/switch/_switch.scss
```diff
@@ -0,0 +1,72 @@
...
+    content: " ";
+    position: absolute;
+    transition: right .2s ease-in-out;
```
sumul: `0.15s ease-out`

library/components/switch/_switch.scss
```diff
@@ -0,0 +1,72 @@
...
+.Switch {
+  @include switch-sizer;
+  // Cancel the default margin for label tags
+  margin: 0;
+  position: relative;
+}
+
+.Switch__indicator {
+  background-color: $gray-300;
+  transition: background-color .2s linear;
```
sumul: `0.35s ease-out`

library/components/switch/_switch.scss
```diff
@@ -0,0 +1,72 @@
+// scss-lint:disable SelectorFormat ElsePlacement StringQuotes
+
+@import '../../../styles/lib/map-deep-get';
```
sumul: Feels like overkill for a simple component to me, but your call.

gelsto: I generally like this pattern for mixins that have sets of sizes, but given that this only has 1 size I'll remove it. 👍

library/components/switch/_switch.scss
```diff
@@ -0,0 +1,64 @@
+// scss-lint:disable SelectorFormat ElsePlacement StringQuotes
+
+@mixin switch-sizer() {
+  $width: 41px;
+  $height: 24px;
```
NickDunkman: Should use modular scale helpers for these:

```sass
$width: modular-scale-px(7);
$height: modular-scale-px(3);
```

library/components/switch/_switch.scss
```diff
@@ -0,0 +1,64 @@
...
+  $toggler-radius: 20px;
+  $toggler-margin: ($height - $toggler-radius) / 2;
+
+  .Switch__indicator {
+    border-radius: ($height / 2);
```
NickDunkman: Can we just use `border-radius: 50%` and move that out of the sizing mixin?

library/components/switch/_switch.scss
```diff
@@ -0,0 +1,64 @@
+// scss-lint:disable SelectorFormat ElsePlacement StringQuotes
+
+@mixin switch-sizer() {
+  $width: 41px;
+  $height: 24px;
+  $toggler-radius: 20px;
+  $toggler-margin: ($height - $toggler-radius) / 2;
```
NickDunkman: The "margin" feels more like a border that will always be our standard border size: `2px`.

Then you can calculate the toggler *diameter* from that, instead of hardcoding a large non-modular scale number in `20px`. That feels like something we generally want to avoid.

```sass
$toggler-diameter: $height - ($border-width * 2);
```

library/components/switch/_switch.scss
```diff
@@ -0,0 +1,64 @@
...
+
+  .Switch__indicator {
+    border-radius: ($height / 2);
+    height: $height;
+    width: $width;
+
+    &::after {
+      height: $toggler-radius;
+      right: $width - $toggler-margin - $toggler-radius;
+      top: $toggler-margin;
```
NickDunkman: if this indeed is just always a border of 2px, we could just do `top: $border-width;` outside of the sizing mixin.

library/components/switch/_switch.scss
gelsto: nah that's a ticket to crazy town:
image

It needs to be half the height.

library/components/switch/switch.jsx
```diff
@@ -0,0 +1,19 @@
+const React = require('react');
+const classes = require('classnames');
+
+function Switch(props) {
+  const { className, isOn, ...inputProps } = props;
+
+  return (
+    <label className={classes(className, 'Switch')}>
+      <input className="Switch__input" type="checkbox" checked={props.isOn} {...inputProps} />
```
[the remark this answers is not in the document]
gelsto: That's a good point, we should just use `checked`. 👍

gelsto:
@NickDunkman @sumul: Everything should be addressed now. Also added the `cursor: not-allowed` for disabled switches as we discussed in slack :)

library/components/switch/README.md
```diff
@@ -0,0 +1,28 @@
...
+### Props
+
+| name | type | default | description |
+| ---- | ---- | ------- | ----------- |
+| `className` | string | | Class name for the root element.
+| `isOn` | boolean | false | If the switch is toggled on. If set to `True`, the underlying `input` element will have a `checked=true` attribute.
+
+Valid HTML attributes and event handlers for the input element (ex. `disabled`, `onChange`) may be safely passed as additional props. **These props will be applied to the `<input>` element, not the root element of the component.**
+
+Passing a prop that is not a legal DOM attribute will cause React to trigger an [unknown prop warning](https://facebook.github.io/react/warnings/unknown-prop.html).
```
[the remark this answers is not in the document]
gelsto: That's fair - maybe we should put this warning in a guide and remove it from the individual READMEs?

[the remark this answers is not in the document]
NickDunkman: I think the warning you get in your console is sufficient ... whenever I see it I feel like I have to fix immediately.

library/components/switch/_switch.scss
NickDunkman: oof

package.json
```diff
@@ -24,7 +24,7 @@
     "webpack": "^2.4.1"
   },
   "scripts": {
-    "test": "mocha --compilers js:babel-register --require tests/test_setup.js --recursive components/**/*.spec.jsx",
+    "test": "mocha --compilers js:babel-register --require tests/test_setup.js --recursive library/components/**/*.spec.jsx",
```
NickDunkman: 😬

library/components/switch/switch.jsx
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <label className={classes(className, 'Switch')}>
+      <input className="Switch__input" type="checkbox" {...inputProps} />
+      <div className="Switch__indicator" />
+    </label>
+  );
+}
+
+Switch.defaultProps = {
+  isOn: false,
```
NickDunkman: Looks like forgot to remove this

library/components/switch/switch.spec.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+const Switch = require('./switch');
+
+describe('Switch', function () {
+  it('should render an "on" state', function () {
+    expect(shallow(<Switch isOn />).find('input').prop('checked')).to.be.true;
+  });
+
+  it('should render an "off" state', function () {
+    expect(shallow(<Switch isOn={false} />).find('input').prop('checked')).to.be.false;
```
NickDunkman: Also here and directly above ... and should have to update these tests to use `checked` prop

PR #51, opened 2017-10-27
Fix all eslint errors

NickDunkman:


.eslintrc
```diff
@@ -10,13 +10,15 @@
...
     "comma-dangle": ["error", "always-multiline"],
+    "function-paren-newline": ["error", "consistent"],
     "func-names": "off",
     "import/no-extraneous-dependencies": "off",
     "import/no-unresolved": "off",
     "jsx-a11y/label-has-for": "off",
     "max-len": ["error", { "code": 120 }],
     "no-unused-expressions": ["error", { "allowShortCircuit": true, "allowTernary": true }],
     "no-underscore-dangle": ["error", { "allow": ["__preview"] }],
+    "object-curly-newline": "off",
```
NickDunkman: As discussed ... we already have a max line length linter, so this just ends up being annoying when destructuring objects.

.eslintrc
```diff
@@ -10,13 +10,15 @@
   "rules": {
     "curly": ["error"],
     "comma-dangle": ["error", "always-multiline"],
+    "function-paren-newline": ["error", "consistent"],
```
NickDunkman: Be default, this linter would raise an error if you had line breaks on a function call with a single argument ... but I think using line breaks in this example is 👍, even though there's only a single arg:

```js
shallow(
  <SomeComponent
    foo="bar"
    blep="mlep"
    fwup="fwop"
  />
);
```

.eslintrc
```diff
@@ -10,13 +10,15 @@
...
     "func-names": "off",
     "import/no-extraneous-dependencies": "off",
     "import/no-unresolved": "off",
     "jsx-a11y/label-has-for": "off",
     "max-len": ["error", { "code": 120 }],
```
NickDunkman: We use 99 max line length in python, should we use the same here? cc @iqnivek

iqnivek: sure, works for me!

library/utilities/sizing/sizing.jsx
```diff
@@ -1,5 +1,5 @@
-import React from 'react';
-import classes from 'classnames';
+const React = require('react');
+const classes = require('classnames');
```
iqnivek: why are we moving away from ES6 imports?

NickDunkman: We were almost exclusively using `require` ... I think to avoid linting errors. @gelsto can you think a reason we shouldn't use `import`?

I updated everything to use `import` and quieted the lint error, and everything seems fine: https://github.com/aminohealth/phenotypes/pull/51/commits/f403a942809d0eb2df6f93519b2e1da60d440c8b

iqnivek: thanks for updating 👍 definitely a case of fixing the linter instead of the code, I think!

gelsto: I think we should use `import`.
We never had a conversation about what ES6 conventions we want to be using in this repo. Generally I think sticking to our Amino conventions will be the easiest for us since it would involve less context-switching for the people who are writing most (all) of the code, but that might increase friction for potential open-source contributors down the line. I think that was the reasoning behind sticking to the default eslint settings for the most part.
We're not using any crazy presets in our Amino build, so I think we should just start using the `pare_web` conventions here. I think that's a separate ticket since it involves some webpack tweaks, though.

library/components/button/anchor_button.jsx
```diff
@@ -23,6 +22,8 @@ class AnchorButton extends React.Component {
 
   render() {
     const { children, disabled, tabIndex, ...htmlProps } = this.props;
+
+    /* eslint-disable jsx-a11y/click-events-have-key-events */
```
gelsto: Should we be keeping this rule?
https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/click-events-have-key-events.md

gelsto: Seems like we never ever follow it....

iqnivek: seems like a relevant article: http://www.karlgroves.com/2014/11/24/ridiculously-easy-trick-for-keyboard-accessibility/

from what I can tell that seems like a slightly painstaking rule to maintain for a relatively niche use case.. while accessibility is important in general, I can go either way on this one.

PR #52, opened 2017-10-30
Add spacebar support to AnchorButton

NickDunkman:
http://10.5.64.52:3000/components/detail/button

Inspired by http://www.karlgroves.com/2014/11/24/ridiculously-easy-trick-for-keyboard-accessibility/

Made enter key & spacebar behavior on `AnchorButton` match `Button`:

- enter presses:
    - this already triggers the click event, so the link will already be followed
    - event happens on key *down* (immediately)
    - the anchor button *does not* get the active style on the button. This is because the click event isn't fired on clicks until you let go of the mouse, but enter causes the event on key *down*
- spacebar presses
    - event does not happen until key *up* (but active style is applied on key *down*)

### DIFF
--- a/library/components/button/anchor_button.jsx
+++ b/library/components/button/anchor_button.jsx
@@ -44,14 +44,12 @@ class AnchorButton extends React.Component {
 
   handleKeyUp(event) {
     if (isSpacebarEvent(event)) {
-      const { disabled, href, onClick } = this.props;
-      if (!disabled) {
+      if (!this.props.disabled) {
         event.preventDefault();
-        if (href && href !== '#') {
-          window.location.href = href;
-        }
-        onClick && onClick(event);
         this.setState({ spacePressed: false });
+        // Follow the link & call onClick prop by simulating a click on the anchor, so that
+        // props like `target` are obeyed.
+        this.element.click();
       }
     }
   }
@@ -62,6 +60,7 @@ class AnchorButton extends React.Component {
     return (
       <a
         {...htmlProps}
+        ref={(element) => { this.element = element }}
         className={classes(className, { 'Button--is-active': this.state.spacePressed })}
         onClick={this.handleAnchorClick}
         onKeyDown={this.handleKeyDown}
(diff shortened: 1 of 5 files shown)


styles/phenotypes.css
```diff
@@ -413,7 +413,8 @@ hr {
   pointer-events: none; }
 
 .Button:active,
-.Button--is-active {
+.Button--is-active,
+.Button--is-active:focus {
```
NickDunkman: The `:focus` selector seems to take precedence over `.Button--is-active`, so needed to make this extra specific selector in order for the `--is-active` style to appear whilst using the spacebar

NickDunkman:
Also: removed the `.Button--is-focused` and `.Button--is-hovered` classes ... after discussing with @gelsto, we don't think those will ever be used.

`.Button--is-disabled` and `.Button--is-active` are needed to make anchor buttons behave like regular buttons, so keeping those.

library/components/button/button.config.js
```diff
@@ -20,8 +20,8 @@ module.exports = {
     name: 'anchor',
     label: 'Using anchor tag',
     context: {
-      href: '#',
-      children: 'Anchor-tagged button',
+      href: 'https://amino.com',
```
gelsto: Small nitpick for usability -- can we add a `target: "__blank"` prop here so we don't redirect the iframe on preview?

NickDunkman:
Updated the spacebar handler to actually trigger a click event on the `<a>` element, so that using spacebar to click the button will still obey things like `target="_blank"`

library/components/button/anchor_button.jsx
```diff
@@ -20,13 +28,43 @@ class AnchorButton extends React.Component {
...
+      if (!this.props.disabled) {
+        event.preventDefault();
+        this.setState({ spacePressed: true });
+      }
+    }
+  }
+
+  handleKeyUp(event) {
+    if (isSpacebarEvent(event)) {
+      if (!this.props.disabled) {
```
gelsto: Since we have the `-1` for tabIndex, do we need to check if it's disabled here?

NickDunkman: Yea - I put it here b/c thought it could still be focused programatically, but I can't actually get that to happen. So will take it off.

gelsto:
✨

PR #53, opened 2017-10-30
Don’t re-render unless actually need to apply active class

NickDunkman:
@gelsto 

Didn't realize that calling setState without any actual changes will call `render` again anyway ... if you hold down spacebar, the keydown event can be triggered many, many times. This will avoid re-rendering many, many times.


gelsto:
👍 seems reasonable to me

PR #54, opened 2017-10-30
Feature/use cursor not allowed for disabled inputs

gelsto:
### Buttons
- Because we use `pointer-events: none` on disabled buttons, both regular and anchor tagged buttons are wrapped with an extra div that manages the cursor. The `a` or `button` tags now have a `Button__control` class that's used to style them (instead of just `Button`, which still goes on the root element).

### Text Inputs
- This was easy, just switched the `cursor` property

### Radio/Checkboxes
- Added an `is-disabled` modifier to the wrapping element that manages the cursor

Just a thought: It seems like we're at a place where all the `is-*` modifiers are classes that should only be applied by the internal component. I don't know if there's a way to indicate a 'private' BEM modifier, but that's essentially the way these should be treated (ex. the user shouldn't be passing `Button--is-disabled` to the button component, since that doesn't change the button behavior and only applies a subset of the disabled styles).

### DIFF
--- a/library/components/button/README.md
+++ b/library/components/button/README.md
@@ -26,7 +26,6 @@ Valid HTML attributes and additional event handlers for the `button` or `a` elem
 | `Button--small` | false | true | |
 | `Button--medium` | true | true | |
 | `Button--large` | false | true | |
-| `Button--is-disabled` | false | false | If the component has a truthy `disabled` prop, this modifier is applied automatically. |
 | `Button--primary` | false | false | |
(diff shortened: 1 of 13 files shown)


NickDunkman:
The privateness of the `is-*` classes feels like a purely react thing ... from CSS's perspective, it's just another old modifier that needs to be applied in the markup. So I think we should just make a note that in the README that React manages that class, and not change the BEM schema.

PR #55, opened 2017-11-02
Feature/progress bar

gelsto:
progress_bar

### DIFF
--- a/library/components/progress_bar/README.md
+++ b/library/components/progress_bar/README.md
@@ -5,15 +5,15 @@
 | `className` | string | | Class name for the root element.
 | `currentIndex` | number | 0 | Zero-based index for the current step.
 | `size` | number | | Total number of steps.
-| `stepClassName` | string | | Class name for individual step elements.
+| `stepClassName` | string | `flex-1` | Class name for individual step elements. Default option expands step elements so the root element will fill its container.
 
 ### Usage
 
-The width for step elements is configured using the `stepClassName` prop. To render a progress bar that expands to the width of its container, pass `stepClassName="flex-1"`.
+The width for step elements is configured using the `stepClassName` prop. For example, to render a progress bar with 36px steps, pass `stepClassName="width-6"`.
 
 Javascript:
 ```html
-<ProgressBar currentIndex={1} size={3} stepClassName="flex-1"/>
+<ProgressBar currentIndex={1} size={3} />
 ```
 Rendered markup:
 ```html
--- a/library/components/progress_bar/README.md
+++ b/library/components/progress_bar/README.md
@@ -1,0 +1,25 @@
+### Props
+
+| name | type | default | description |
+| ---- | ---- | ------- | ----------- |
+| `className` | string | | Class name for the root element.
+| `currentIndex` | number | 0 | Zero-based index for the current step.
+| `size` | number | | Total number of steps.
+| `stepClassName` | string | | Class name for individual step elements.
+
+### Usage
+
+The width for step elements is configured using the `stepClassName` prop. To render a progress bar that expands to the width of its container, pass `stepClassName="flex-1"`.
+
+Javascript:
+```html
+<ProgressBar currentIndex={1} size={3} stepClassName="flex-1"/>
+```
+Rendered markup:
+```html
+<ul class="ProgressBar">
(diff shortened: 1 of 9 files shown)


library/components/progress_bar/_progress_bar.scss
```diff
@@ -0,0 +1,38 @@
...
+  margin-right: $border-width * 1.5;
+  position: relative;
+
+  &:last-child {
+    margin-right: 0;
+  }
+
+  &::after {
+    background-color: $purple-200;
+    box-shadow: 0 0 modular-scale-px(-6) 0 $purple-100;
```
gelsto: @sumul wanted to call out that the `box-shadow` color in the sketch file isn't really `$purple-100`, but it's close so I used it here. Can use `#D89FDF` if that was intentional, though.

library/components/progress_bar/README.md
```diff
@@ -0,0 +1,25 @@
...
+| name | type | default | description |
+| ---- | ---- | ------- | ----------- |
+| `className` | string | | Class name for the root element.
+| `currentIndex` | number | 0 | Zero-based index for the current step.
+| `size` | number | | Total number of steps.
+| `stepClassName` | string | | Class name for individual step elements.
+
+### Usage
+
+The width for step elements is configured using the `stepClassName` prop. To render a progress bar that expands to the width of its container, pass `stepClassName="flex-1"`.
```
NickDunkman: Think it would be simpler to have the progress bars *always* expand to fill the container, and then just tinker with the container's width? That seems easier to reason about 🤔

library/components/progress_bar/_progress_bar.scss
sumul: Oh, good catch. I think it has something to do with using the color picker eye dropper thing in Sketch on this wide gamut screen and Sketch's color management. It's meant to be `$purple-100` 💯.

library/components/progress_bar/README.md
sumul: I think it would be a nice default to fill the container (i.e. `flex-1` is implied and you never have to remember that), but I'd still want to be able to have specifically sized steps as an option. 

This might be a good time to add Phenotypes JS tokens for things like spacing, sizing, and color values. IMO a nice syntax for this component could look something like:

```
import {sizes} from 'utils.js';

// Full width (default)
<ProgressBar currentIndex={1} numSteps={3} />

// Specifically-sized bars
<ProgressBar currentIndex={1} numSteps={3} stepWidth={sizes.size4} />
```

library/components/progress_bar/_progress_bar.scss
```diff
@@ -0,0 +1,38 @@
...
+
+.ProgressBar__step,
+.ProgressBar__step::after {
+  border-radius: $border-radius;
+  display: inline-block;
+}
+
+.ProgressBar__step {
+  background-color: rgba($black, $alpha-300);
+  height: $border-width;
```
sumul: This isn't really meant to be conceptually coupled to the default border width. I could imagine wanting to change the border with to 1px but not wanting to affect progress bars in any way. Let's just hardcode this as 2px and the margin as 3px.

library/components/progress_bar/README.md
sumul: Hmm, although I'm not sure how we'd get the responsive behavior this way. Maybe your approach is the way to go.

gelsto: If we want a way to have specifically sized steps, I think I like the option to make them responsive. I added `flex-1` as the default class here: c8a0e11.

library/components/progress_bar/progress_bar.jsx
```diff
@@ -0,0 +1,25 @@
...
+            'ProgressBar__step--active': index <= currentIndex,
+          })}
+        />
+      ))}
+    </ul>
+  );
+}
+
+ProgressBar.defaultProps = {
+  currentIndex: 0,
```
NickDunkman: The use of the word `index` is slightly confusing to me, b/c usually the first item's index is 0, and thus it's a little ambiguous whether passing zero actually highlights the first step or not. How about `progress`?

```jsx
<ProgressBars progress={0} /> // no progress so far
<ProgressBars progress={1} /> // one bar finished
```

PR #56, opened 2017-11-03
Slider component

NickDunkman:
@sumul I didn't add any color modifiers yet - do you think we should enshrine the red & orange in the HSA designs? Or are those one offs that we should just add custom modifiers for in pare_web?

I spent some time trying to figure out why zeros in our component config files get changed to null - but it doesn't seem to be anything we can fix in this repository. So instead of specifying variants in the config file and then changing nulls back to 0 in the JS, I'm just doing everything in the demo jsx file.

slider

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,6 +10,7 @@
   "rules": {
     "curly": ["error"],
     "comma-dangle": ["error", "always-multiline"],
+    "default-case": "off",
     "function-paren-newline": ["error", "consistent"],
     "func-names": "off",
     "import/extensions": "off",
--- a/library/components/slider/_slider.jsx
+++ b/library/components/slider/_slider.jsx
@@ -1,26 +1,36 @@
 import React from 'react';
 import Slider from './slider.jsx';
 
+module.exports = function demos() {
+  return (
+    <div>
+      <SliderDemo value={50} min={0} max={100} step={1} />
+      <SliderDemo value={1.75} min={1} max={2} step={0.01} />
+      <SliderDemo value={25} min={0} max={100} step={1} disabled />
+    </div>
+  );
+};
+
 class SliderDemo extends React.Component {
   constructor(props, context) {
     super(props, context);
-    this.state = { value: 50 };
+    this.state = { value: props.value };
   }
 
   render() {
+    const props = Object.assign({}, this.props);
+    props.value = this.state.value;
+
     return (
-      <div>
-        <h6>{this.state.value}</h6>
+      <div className="mb7">
         <Slider
-          value={this.state.value}
-          step={0.1}
-          min={49}
-          max={51}
-          onChange={(val) => { this.setState({ value: val }) }}
+          {...props}
+          onChange={(newValue) => { this.setState({ value: newValue }) }}
         />
+        <div className="mt1 text-color-secondary text-2">
+          <code>{JSON.stringify(props)}</code>
+        </div>
       </div>
     );
   }
 }
-
-module.exports = SliderDemo;
--- a/library/components/slider/_slider.jsx
+++ b/library/components/slider/_slider.jsx
@@ -1,0 +1,10 @@
+import React from 'react';
+import Slider from './slider.jsx';
+
+function SliderDemo() {
+  return (
+    <Slider value={54} />
+  );
+}
+
+module.exports = SliderDemo;
--- a/library/components/slider/_slider.jsx
+++ b/library/components/slider/_slider.jsx
@@ -18,8 +18,7 @@ class SliderDemo extends React.Component {
   }
 
   render() {
-    const props = Object.assign({}, this.props);
-    props.value = this.state.value;
+    const props = { ...this.props, value: this.state.value };
 
     return (
       <div className="mb7">
--- a/library/components/slider/_slider.jsx
+++ b/library/components/slider/_slider.jsx
@@ -1,10 +1,26 @@
 import React from 'react';
 import Slider from './slider.jsx';
 
-function SliderDemo() {
-  return (
-    <Slider value={54} />
-  );
+class SliderDemo extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+    this.state = { value: 50 };
+  }
+
+  render() {
+    return (
+      <div>
+        <h6>{this.state.value}</h6>
+        <Slider
+          value={this.state.value}
+          step={0.1}
+          min={49}
+          max={51}
+          onChange={(val) => { this.setState({ value: val }) }}
+        />
+      </div>
+    );
+  }
 }
 
 module.exports = SliderDemo;
--- a/library/components/slider/_slider.scss
+++ b/library/components/slider/_slider.scss
@@ -43,6 +43,11 @@ $slider-track-border-radius: $slider-track-height / 2;
   width: $slider-handle-size;
 }
 
+.Slider--is-disabled {
+  cursor: not-allowed;
+  opacity: $alpha-500;
+}
+
 @mixin slider-color($fill-color, $track-color: rgba($black, $alpha-300), $handle-color: $white) {
   .Slider__track {
     background-color: $track-color;
@@ -60,9 +65,5 @@ $slider-track-border-radius: $slider-track-height / 2;
 @mixin slider-green {
   @include slider-color($green-200);
 }
-
-//@mixin slider-reversed {
-//  @include slider-color(rgba($white, $alpha-reversed-600), rgba($white, $alpha-reversed-200));
-//}
 
 @include slider-green;
--- a/library/components/slider/_slider.scss
+++ b/library/components/slider/_slider.scss
@@ -49,6 +49,20 @@ $slider-focus-color: $blue-100;
   right: -1 * $slider-handle-size / 2;
   top: -1 * $slider-track-offset;
   width: $slider-handle-size;
+
+  &::after {
+    background-color: $slider-focus-color;
+    border-radius: 50%;
+    bottom: $slider-focus-margin;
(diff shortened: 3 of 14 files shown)


library/components/slider/slider.jsx
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  handleDrag(event) {
+    if (this.waitingForDragAnimationFrame) {
+      return;
+    }
+
+    this.waitingForDragAnimationFrame = true;
+
+    requestAnimationFrame(() => {
```
NickDunkman: took this trick from material-ui's implementation - basically it throttles the drag event down to the speed at which the browser does repaints

.eslintrc
```diff
@@ -10,6 +10,7 @@
   "rules": {
     "curly": ["error"],
     "comma-dangle": ["error", "always-multiline"],
+    "default-case": "off",
```
NickDunkman: I think this rule is dumb ... by default you get a lint error if you have a `switch` statement that does not include a `default` case. But you shouldn't have to include one if you don't want to have any logic for that case ...

library/components/slider/slider.jsx
```diff
@@ -0,0 +1,258 @@
...
+      <div
+        role="slider"
+        aria-valuemax={max}
+        aria-valuemin={min}
+        aria-valuenow={value}
+        {...other}
+        ref={(element) => { this.slider = element }}
+        className={classes('Slider', className, {
+          'Slider--is-disabled': disabled,
+          'Slider--is-focused': !disabled && this.state.useFocusStyle,
```
gelsto: (I'm biased because I wrote the switch this way, but) I think it would be clearer if you managed the focus state by:
```es6
handleKeyDown() {
  /* ... */
  this.setState({ clicked: false });
}

handleMouseDown() {
  /* ... */
  this.setState({ clicked: true });
}

handleFocus(event) {
  this.setState({ focused: true })
  this.props.onFocus && this.prop.onFocus(event);
}

handleBlur(event) {
  this.setState({ focused: false, clicked: false })
  this.props.onBlur && this.prop.onBlur(event);
}

render() {
  return (
     <div
       role="slider"
       /* ... */
      className={classes('Slider', className, {
        'Slider--is-disabled': disabled,
        'Slider--is-focused': !disabled && this.state.focused && !this.state.clicked,
      })}
     /* ... */
 );

}
```
You have to keep track of 2 things anyway, and I think `focused` and `clicked` are more obvious and easier to follow than `useFocusStyle` and `focusedByClick`

library/components/slider/slider.jsx
```diff
@@ -0,0 +1,258 @@
...
+  setValueFromEvent(event) {
+    const { min, max, step, value, onChange } = this.props;
+    const track = this.track.getBoundingClientRect();
+
+    let eventX = event.touches ? event.touches[0].clientX : event.clientX;
+    if (eventX < track.left) {
+      eventX = track.left;
+    } else if (eventX > track.right) {
+      eventX = track.right;
+    }
```
iqnivek: This handles the fact that there's a gutter on the left and right of the track with width of the slider handle's radius, right? Might be nice to add a small comment explaining the logic here

library/components/slider/slider.jsx
iqnivek: cool, TIL

gelsto: Would we want to think about extracting this into a utility in the future?

library/components/slider/slider.jsx
```diff
@@ -0,0 +1,258 @@
...
+    this.props.onDragEnd && this.props.onDragEnd(event);
+  }
+
+  handleTouchStart(event) {
+    const { onTouchStart, onDragStart } = this.props;
+
+    document.addEventListener('touchmove', this.handleDrag);
+    document.addEventListener('touchup', this.handleTouchEnd);
+    document.addEventListener('touchend', this.handleTouchEnd);
+    document.addEventListener('touchcancel', this.handleTouchEnd);
```
gelsto: I'm not against this if it doesn't cause performance issues, but I'm curious why we're adding/removing the listeners every time?

NickDunkman: Putting these events on the document gives you more freedom when dragging the mouse/finger, but then we need to take them off when you're done dragging, so that they don't fire when you just click/touch anywhere on the page (and then change our value & call our onDragStart/etc. callbacks).

library/components/slider/slider.jsx
```diff
@@ -0,0 +1,258 @@
...
+    this.state = { useFocusStyle: false };
+
+    this.handleDrag = this.handleDrag.bind(this);
+    this.handleMouseDown = this.handleMouseDown.bind(this);
+    this.handleDragMouseEnd = this.handleDragMouseEnd.bind(this);
+    this.handleTouchStart = this.handleTouchStart.bind(this);
+    this.handleTouchEnd = this.handleTouchEnd.bind(this);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.handleFocus = this.handleFocus.bind(this);
+    this.handleBlur = this.handleBlur.bind(this);
```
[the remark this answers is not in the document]
NickDunkman: Yea i agree this would be really nice - the babel/preset stuff is a little over my head still.

library/components/slider/slider.jsx
```diff
@@ -0,0 +1,258 @@
...
+
+    onKeyDown && onKeyDown(event);
+
+    // Prevent all the keys that interact with the slider from moving the page around
+    event.preventDefault();
+  }
+
+  handleFocus(event) {
+    if (!this.focusedFromClick) {
+      this.focusedFromClick = false;
```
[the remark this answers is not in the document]
iqnivek: agreed

library/components/slider/slider.jsx
NickDunkman: https://github.com/aminohealth/phenotypes/pull/56/commits/d3b936063895eafb6bc5fedc0914db073e325c43

library/components/slider/slider.jsx
```diff
@@ -0,0 +1,258 @@
...
+Slider.defaultProps = {
+  disabled: false,
+  max: 100,
+  min: 0,
+  step: 1,
+  tabIndex: 0,
+  value: null,
+};
+
+module.exports = Slider;
```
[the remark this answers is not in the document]
NickDunkman: Yea that does work here ... i've seen errors with using `export` in some places in the repo - haven't pinpointed at what level of Fractal that occurs. Will follow up and change to `export` soon.

library/components/slider/slider.jsx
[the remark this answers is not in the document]
NickDunkman: I don't like how that would result in a re-render ... gonna try to figure out how to wrap that in & the reloadAnimationFrame stuff in a closure in a helper function ...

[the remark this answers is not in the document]
iqnivek: does the re-render cause a visible performance issue? if not, it might not be worth worrying about too much. React re-renders stuff all the time, and I don't expect the handleFocus callback to be called particularly frequently.

library/components/slider/_slider.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+class SliderDemo extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+    this.state = { value: props.value };
+  }
+
+  render() {
+    const props = Object.assign({}, this.props);
+    props.value = this.state.value;
```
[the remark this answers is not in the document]
NickDunkman: yas! that's nice. https://github.com/aminohealth/phenotypes/pull/56/commits/371ba62cb8267c37ce63e5d818b4959050cf48c4

library/components/slider/slider.jsx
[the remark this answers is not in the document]
NickDunkman: got it into a helper 💪 : https://github.com/aminohealth/phenotypes/pull/56/commits/1bdce6eac331d117aa4f1bcc4b81a87fcdc2ab20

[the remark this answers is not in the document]
NickDunkman: Oh whoops - was thinking about something else. Ignore comment above.

library/components/slider/slider.jsx
NickDunkman: https://github.com/aminohealth/phenotypes/pull/56/commits/1bdce6eac331d117aa4f1bcc4b81a87fcdc2ab20

library/components/slider/slider.jsx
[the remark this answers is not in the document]
sumul: @iqnivek I defer to you and @gelsto on that.

library/components/slider/_slider.scss
```diff
@@ -0,0 +1,91 @@
...
+  width: $slider-handle-size;
+}
+
+.Slider--is-disabled {
+  cursor: not-allowed;
+  opacity: $alpha-500;
+}
+
+.Slider--is-focused {
+  .Slider__handle::after {
```
sumul: Would be nice to match the transition that @gelsto used on the switch focus indicator.

NickDunkman: https://github.com/aminohealth/phenotypes/pull/56/commits/3f54e9b0b48746f25173c39a12498b908da02177

library/components/slider/slider.jsx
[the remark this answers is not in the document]
gelsto: @iqnivek @NickDunkman there's a ticket for this here: https://www.pivotaltracker.com/story/show/152410693

Agreed on the stage-2 vs. stage-0 concern. We're not using any stage-0 conventions in pare_web even though we're set up to do so. I believe there was a ticket *somewhere* to just change it to stage-2, but I'm not sure which backlog it's in.

PR #64, opened 2017-12-12
Font antialiasing and text-color utilities

NickDunkman:
- adds `-webkit-font-smoothing: antialiasing` and `-moz-osx-font-smoothing: grayscale;` (for firefox) to the `<body>`. The non-vendor-prefixed `font-smooth` property doesn't appear to do anything (I tried in all the main browsers), so left that out.
- adds text color utilities, with docs

### breaking changes

- the `reversed-text` mixin has been removed (it was mainly adding antialiasing for reversed text, which is no longer necessary)
- antialiasing of text is kinda breaking? since the style of all text is slightly changing

### DIFF
--- a/library/components/button/_button.scss
+++ b/library/components/button/_button.scss
@@ -90,8 +90,9 @@
 
 .Button--danger {
   .Button__control {
-    @include reversed-text;
     background: $red;
+    color: $white;
+    font-weight: $font-weight-bold;
 
     &:hover,
     &:focus {
@@ -112,8 +113,9 @@
 .Button--primary {
   .Button__control {
     @include depth(4);
-    @include reversed-text($font-weight: $font-weight-bold);
     background: $blue;
+    color: $white;
+    font-weight: $font-weight-bold;
 
     &:hover,
     &:focus {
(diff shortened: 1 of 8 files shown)


library/components/button/_button.scss
```diff
@@ -90,8 +90,9 @@
 
 .Button--danger {
   .Button__control {
-    @include reversed-text;
     background: $red;
+    color: $white;
+    font-weight: $font-weight-bold;
```
NickDunkman: `.Button--danger` now needs bold weight (previously the non-antialiased version looked better as normal weight, but with antialiasing, the bold weight is better).

iqnivek:
> antialiasing of text is kinda breaking? since the style of all text is slightly changing

So I think just as a clarification of "breaking change" as it relates to semver - I think if a style changes but doesn't require the end-user to update any code, then it doesn't call for a semver major version update.

Just wanted to bring this up because I think this may have been the case for the grid padding change which we major-version'd into 4.0.0 - since that grid change does change the styles but is also something that pare_web didn't need to make a code change for, I think that could've been a patch version change instead of major.

Of course, removal of CSS does qualify as a breaking change. But if it were just the antialiasing/utility additions, I think this PR could be a minor version update.

NickDunkman:
@sumul Fixed the weird char issue & also bumped to 5.0.0

PR #66, opened 2017-12-21
Add support for setting the width of buttons with our normal sizing classes

ewdicus:
This adds `width: 100%` to the `Button__control` class so the inner button or anchor element will expand to the size of its wrapping `Button` div.

We could consider adding something like `height: inherit` to `Button__control` as well, but I don't see any reason we'd set the height outside of the button sizing classes currently.

Also updated the tests and the button docs usage section.

### DIFF
--- a/library/components/button/README.md
+++ b/library/components/button/README.md
@@ -41,6 +41,10 @@ Javascript:
 ```
 Rendered markup:
 ```html
-<button class="Button" type="button">Click me!</button>
-<a href="#" class="Button" role="button">No, click me!</a>
+<div class="Button">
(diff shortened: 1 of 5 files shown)


library/components/button/README.md
```diff
@@ -41,6 +41,10 @@ Javascript:
 ```
 Rendered markup:
 ```html
-<button class="Button" type="button">Click me!</button>
-<a href="#" class="Button" role="button">No, click me!</a>
+<div class="Button">
+    <button class="Button__control" type="button">Click me!</button>
```
iqnivek: indent two spaces

sumul:
@iqnivek Sounds good!

PR #73, opened 2019-01-07
Make compatible with Babel 7

NickDunkman:
Since we don't build phenotypes here yet, & consumers of phenotypes import raw source directly, there are a couple things we need to do to make compatible with a consumer that is running Babel 7:

- consistently use commonjs or es6 exports/imports (I chose es6)
- update dependencies to be compatible with Babel 7


package.json
```diff
@@ -3,29 +3,31 @@
   "version": "6.0.0",
   "description": "Amino's design language and front-end component library",
   "dependencies": {
+    "@babel/register": "^7.0.0",
```
gelsto: Should this be a dev dependency?

NickDunkman: Yep! https://github.com/aminohealth/phenotypes/pull/73/commits/155c9f3cbdf66a1fc0fa198ffb5a448001f0b357

PR #76, opened 2019-06-05
Get fstream 1.0.12 as node-sass dep

gelsto:
See https://github.com/sass/node-sass/issues/2676

Our node-sass version is the same, I just uninstalled/reinstalled it at the pinned version

### DIFF
--- a/styles/phenotypes.css
+++ b/styles/phenotypes.css
@@ -7,30 +7,30 @@
 /**
  * @license
  * MyFonts Webfont Build ID 3386717, 2017-05-09T18:37:44-0400
- * 
+ *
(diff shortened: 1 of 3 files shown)


styles/phenotypes.css
```diff
@@ -7,30 +7,30 @@
 /**
  * @license
  * MyFonts Webfont Build ID 3386717, 2017-05-09T18:37:44-0400
- *
```
sumul: @NickDunkman Hope you're not too bummed that all these trailing spaces are going to be missing now.

NickDunkman: :trollface:

PR #85, opened 2020-01-28
Add more sizes (10, 11, 12, 13)

NickDunkman:
I want to be able to do this in web:

```jsx
<div className="height-12 height-auto-sm">
```

But that doesn't work, because `height-12` is defined in web, so it's more specific than `height-auto-sm` which is defined in phenotypes and thus comes earlier in the stylesheet.

I'm only adding these sizes to our height/width utilities (not our margin, padding, etc. utilities), so it's not a huge addition.

### DIFF
--- a/styles/phenotypes.css
+++ b/styles/phenotypes.css
@@ -2936,6 +2936,18 @@ hr {
 .width-9 {
   width: 119px !important; }
 
+.width-10 {
+  width: 177px !important; }
+
+.width-11 {
+  width: 264px !important; }
+
+.width-12 {
+  width: 394px !important; }
+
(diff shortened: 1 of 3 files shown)


styles/phenotypes.css
```diff
@@ -2936,6 +2936,18 @@ hr {
 .width-9 {
   width: 119px !important; }
 
+.width-10 {
+  width: 177px !important; }
```
gelsto: Why don't we use the modular scale mixin here?

NickDunkman: This is the built .css file 😄

PR #88, opened 2020-04-24
Upgrade to node 12 and bump node-sass dependency

cjduffett:
@sumul @NickDunkman this PR upgrades the base docker image used to build this project from `node:boron` to `node:12.6.2-stretch-slim`. `node:boron` appears to be several years out-of-date. We also plan to upgrade pare_web's `www_js` container to that same base image for node 12.

 node-sass `4.12+` is required to use node 12, so I upgraded node-sass from `4.5.3` to `4.13.1`. The latest release is `4.14` but that was published just this week, so I figured `4.13.1` would be a little more stable.

After making the dependency changes I was able to build, test, and run the `fractal` container just fine. Is there anything else I can do to validate that these upgrades don't cause any issues downstream?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:boron
+FROM node:12.16.2-stretch-slim
 
 ENV NPM_CONFIG_LOGLEVEL warn
 
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "dependencies": {
     "classnames": "^2.2.5",
     "keycode": "^2.1.9",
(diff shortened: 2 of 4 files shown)


sumul:
@cjduffett Thanks for taking care of this. I just looked at the diff for phenotypes.css, which is a likely place where things would look obviously wrong if a new version of node-sass were to break things. Nothing noteworthy there, so this should be all good. 

@NickDunkman There have been some security vulnerability notices lately for this repo. Might be worth resolving those too before publishing this as a maintenance release to NPM.

PR #93, opened 2020-06-24
Consolidate library files to a dist directory

ahmedmodan:
This pr consolidates built library files to the dist directory. This includes all of the precompiled css and webfonts. I think it makes sense to leave the scss files in the styles directory as they are not being changed at all when they are exposed for external consumption. 

This pr also changes the css npm scripts to not use the node-sass cli but now runs a node.js script which uses the node-sass api directly. This should allow us a bit more control over running node-sass and should also allow us to easily extend the build in the future when we need to compile two different versions of the library (one with css variables included and one without). 

The `build-css` script doesn't do anything different than what the CLI used to do. The `watch-css` script is pretty different than the previous cli usage though. The previous `watch-css` script used `node-sass` in watch mode to watch the `/styles` directory and rebuild anytime something changed. `node-sass` under the hood would create a sass-graph that would implicitly included all of the scss files that were also in the `/library` tree and were imported into the main scss file. 

But because `node-sass` only has the ability to watch directories we had at least 2-3 rebuilds each time a file changed. 1 when the file we wanted to change changed, (ex. `_button.scss`) and then because the compiled `phenotypes.css` file in the `/styles` directory changed, it would rebuild again, making it two. And then because it would detect that the `phenotypes.css` file has changed again it would rebuild again for good measure. 

On top of that the file watcher that [node-sass currently uses](https://github.com/sass/node-sass/blob/master/bin/node-sass#L5) has serious memory issues when run in a docker container, which is exactly what we do. Node-sass itself is planning to switch to `[chokidar](https://github.com/paulmillr/chokidar)` for v5. Rather than wait for node-sass to release v5 we can use `chokidar` directly to explicitly watch the files we need and rebuild on change.

While it may look like these changes change everything about our build, everyday usage should feel and function exactly the same as it currently does. The only difference being that we shouldn't have insanely high memory usage when running our dev builds in a docker container.


package.json
```diff
@@ -48,7 +50,8 @@
     "start": "fractal start --sync & npm run watch-all",
     "build-library": "rollup --config ./configs/rollup.phenotypes-library.config.js",
     "watch-library": "rollup --config ./configs/rollup.phenotypes-library.config.js --watch",
-    "prepublish": "npm run build-library"
+    "copyAssets": "node scripts/copyLibraryAssets.js",
+    "prepublishOnly": "npm run build-library && npm run build-css && npm run copyAssets"
```
ahmedmodan, line 54: change from `prepublish` to `prepublishOnly` so this script does not run on initial `yarn/npm install`

package.json
```diff
@@ -61,5 +64,7 @@
   },
   "homepage": "https://github.com/aminohealth/phenotypes#readme",
   "module": "./dist/phenotypes.esm.js",
-  "main": "./dist/phenotypes.cjs.js"
+  "main": "./dist/phenotypes.cjs.js",
+  "style": "./dist/css/phenotypes.css",
+  "scss": "./styles/phenotypes.scss"
```
ahmedmodan, line 69: This aligns with how major libraries like [bootstrap](https://github.com/twbs/bootstrap/blob/main/package.json#L74-L75) and [materialize](https://github.com/Dogfalo/materialize/blob/v1-dev/package.json#L8-L9) handle these fields in their package.json

package.json
```diff
@@ -61,5 +64,7 @@
   },
   "homepage": "https://github.com/aminohealth/phenotypes#readme",
   "module": "./dist/phenotypes.esm.js",
-  "main": "./dist/phenotypes.cjs.js"
+  "main": "./dist/phenotypes.cjs.js",
+  "style": "./dist/css/phenotypes.css",
+  "sass": "./styles/phenotypes.scss"
```
ahmedmodan, line 69: This aligns with how major libraries like [bootstrap](https://github.com/twbs/bootstrap/blob/main/package.json#L74-L75) and [materialize](https://github.com/Dogfalo/materialize/blob/v1-dev/package.json#L8-L9) handle these fields in their package.json

package.json
gelsto: Not a blocker, but having camel case _and_ kebab case commands is making me 😭 . Can we just use camel?

ahmedmodan: consistency in script naming! https://github.com/aminohealth/phenotypes/pull/93/commits/3614c56681feea1a77ffe6c3eddbabb110a0dd44

PR #95, opened 2020-06-30
Provide public JS API for component library + upgrade base image to node14

ahmedmodan:
Upgrade the base image from node 8 to node 14. Node 14 is current and will be active LTS until Oct 2021 That should give us plenty of breathing room. This also allows us to update node-sass to the current version which will unblock us on updating the pare_web base image off of node 8.

This release also includes the new JS API for all of the phenotypes react components.

Once published each component will now be available through named imports like so:

```js
import {Button, ProgressBar} from '@aminohealth/phenotypes'
```


ahmedmodan:
After this is merged we can release a new version of phenotypes.

PR #97, opened 2020-07-02
Add initial support for theming via variables

ahmedmodan:
looks like a lot, but if you ignore the generated css files that we are checking in (of which there are two, both at 6177 lines) there around 350 lines worth of changes in here.

Changes included in this pr:
* Adds a `_theme_config.scss` file which contains a `:root` rule containing all of the themable variable declarations
* Updates the build script to compile two different css files, one with css variables included and one without. (this adds a postcss dependency to the build. post-css is used to strip away the css variables for the non-variable build)
* Updates the react components to use the theme variables
* Updates a few utility classes to use the theme variables
* Updates the phenotypes fractal website to use the themable css file. (This breaks IE11 compatibility)
* Updates the phenotypes fractal website to include a `Theming` guide.
* Updates the README.md to provide instructions on theming.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ $grid-breakpoints: (
 
 The `_mixins.scss` file contains imports for functions and mixins that are used by other modules and components. Actual implementations should go in `styles/mixins`. E.g. `styles/mixins/_type.scss` contains both functions and mixins that help with typography.
 
-The `_theme_config.scss` file contains a css ruleset that defines the theming css variables that phenotypes uses. These are defined as css custom properties and allow custom themes for Phenotypes. Please see the [Theming README](/THEMING.md) for more info
+The `_theme_config.scss` file contains a css ruleset that defines the theming css variables that phenotypes uses. These are defined as css custom properties and allow custom themes for Phenotypes. Please see the [Theming guide](/guides/11-theming.md) for more info
 
 Phenotypes uses a slightly modified version of Bootstrap 4's reboot. [Read more about that.](https://v4-alpha.getbootstrap.com/content/reboot/)
 
@@ -152,11 +152,11 @@ $ npm install @aminohealth/phenotypes --save
 
 The recommended way to use the styles for phenotypes is to import the scss entrypoint phenotypes.scss in your own project. When installed via npm this will be located at `@aminohealth/phenotypes/styles/phenotypes.scss`. This does require your project to use [Sass](https://sass-lang.com/). This is; however, not required and for that reason we also provide plain old css that you are able to use.
 
-**NOTE**  By default the phenotypes.scss entry point when compiled will include CSS variables in the output. CSS variables are required for custom themes. CSS variables will work fine for all evergreen browsers; however, if you need to support a browser that does not support CSS variables (IE) or do not require a custom theme, you can compile the css variables away (Or see the [theming docs](./THEMING.md) for supporting IE .) CSS variables can be compiled away easily by post-processing via [postcss](https://postcss.org/) and using the [postcss-preset-env](https://preset-env.cssdb.org/) plugin. An example of how to do this is can be seen in our own build process when we compile the phenotypes.scss file to plain old css [here](./scripts/compileScss.js#L40-L48).
+**NOTE**  By default the phenotypes.scss entry point when compiled will include CSS variables in the output. CSS variables are required for custom themes. CSS variables will work fine for all evergreen browsers; however, if you need to support a browser that does not support CSS variables (IE) or do not require a custom theme, you can compile the css variables away (Or see the [Theming guide](/guides/11-theming.md) for supporting IE .) CSS variables can be compiled away easily by post-processing via [postcss](https://postcss.org/) and using the [postcss-preset-env](https://preset-env.cssdb.org/) plugin. An example of how to do this is can be seen in our own build process when we compile the phenotypes.scss file to plain old css [here](./scripts/compileScss.js#L40-L48).
 
 ### CSS
 
-If you do not want to use sass there are two different css files that Phenotypes provides. The most straightforward of which is located at `@aminohealth/dist/css/phenotypes.css` when installed via npm. The other file is a themable css file that is located at `@aminohealth/dist/css/phenotypes.themable.css`. This file uses CSS variables to set theme variables and propagate those values to the Phenotypes styling. More information on this can be found in the [theming docs](./THEMING.md).
+If you do not want to use sass there are two different css files that Phenotypes provides. The most straightforward of which is located at `@aminohealth/dist/css/phenotypes.css` when installed via npm. The other file is a themable css file that is located at `@aminohealth/dist/css/phenotypes.themable.css`. This file uses CSS variables to set theme variables and propagate those values to the Phenotypes styling. More information on this can be found in the [Theming guide](/guides/11-theming.md).
 
 ### Peer Dependencies
 
@@ -183,7 +183,7 @@ ReactDOM.render(
 
 ## Theming
 
-Phenotypes allows you to theme the css and component library to suit your needs. Please see the [theming docs](/THEMING.md) for more information on how to customize and theme phenotypes.
+Phenotypes allows you to theme the css and component library to suit your needs. Please see the [Theming guide](/guides/11-theming.md) for more information on how to customize and theme phenotypes.
 
 [npm-url]: https://www.npmjs.com/package/@aminohealth/phenotypes
 [npm-version-image]: https://img.shields.io/npm/v/@aminohealth/phenotypes.svg?style=flat
--- a/README.md
+++ b/README.md
@@ -70,10 +70,11 @@ The SASS file organization scheme is loosely based on Bootstrap 4. The entry poi
 2. Feature toggles
 3. Modular scale
 4. Variables and mixins
-5. Reboot
-6. Typography
-7. Components
-8. Utilities
+5. Theme CSS variables
+6. Reboot
+7. Typography
+8. Components
+9. Utilities
 
 Webfonts are imported first, because they're pulled in as a raw .css import. The webfont file has a CSS `@import` directive at the top, which must come first in the compiled stylesheet.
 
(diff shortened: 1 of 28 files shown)


library/components/progress_bar/_progress_bar.scss
```diff
@@ -23,7 +23,7 @@
   }
 
   &::after {
-    background-color: $purple-200;
+    background-color: var(--step-progress-active-color);
     box-shadow: 0 0 modular-scale-px(-6) 0 $purple-100;
```
sumul, line 27: Let's make a variable for the shadow and have it by default be the same as `--step-progress-active-color` at 54% opacity.

styles/_theme_config.scss
```diff
@@ -0,0 +1,37 @@
...
+  --text-color-secondary-reversed: #{$text-color-reversed-secondary};
+  --text-color-hint-reversed: #{$text-color-reversed-hint};
+
+  --brand-color-primary: #{$brand-primary};
+  --brand-color-accent: #{$orange-200};
+  --positive-color: #{$green-200};
+  --negative-color: #{$brand-danger};
+  --warning-color: #{$brand-warning};
+  --error-color: var(--negative-color);
+  --interactive-color: #{$blue-200};
```
sumul, line 16: These seem like great candidates for utility classes.

library/components/progress_bar/_progress_bar.scss
ahmedmodan: @sumul @gelsto @ewdicus I want some input on the following: In this particular commit: https://github.com/aminohealth/phenotypes/pull/97/commits/61f9a1ac5c85128749d0cd62cf23e07d07c8176a i changed how the theme_config variables were created. I added a scss mixin called `color-variable` that creates two different css variables.  
https://github.com/aminohealth/phenotypes/blob/61f9a1ac5c85128749d0cd62cf23e07d07c8176a/styles/mixins/_color.scss#L13-L42

This was to mainly ensure that we could apply alpha channels to variables. It adds two different variables whenever it is used. One that is the variable that we ultimately want to create and one that creates an rgb variable that we can reference, when we want to apply alpha channels. here is an example:

```scss
:root {
  --step-progress-active-color: rgb(179, 93, 186);
}

.ProgressBar__step {
  // ...
  &::after {
    background-color: var(--step-progress-active-color);
    box-shadow: 0 0 modular-scale-px(-6) 9
      rgba(var(--step-progress-active-color), $alpha-600);
      // the above variable usage is not possible since css variables are direct
      // string substitutions. the above will result in css that looks like this:
      // rgba(rgb(179, 93, 186), .54);
  }
}


// in order to fix this we can extract the actual rgb values into a separate
// variable like so:
:root {
  --step-progress-active-color-rgb: 179, 93, 186;
  // we can then in turn use this variable as the value to
  // the variable we want to create.
  --step-progress-active-color: var(--step-progress-active-color-rgb);
}

// This also allows us to use that rgb value again for something like a
// box shadow. Here is the same ProgressBar__step rule as above, except
// it uses the rgb varable for the box shadow. This one is indeed valid css
.ProgressBar__step {
  // ...
  &::after {
    background-color: var(--step-progress-active-color);
    box-shadow: 0 0 modular-scale-px(-6) 9
      rgba(var(--step-progress-active-color-rgb), $alpha-600);
  }
}
```

I want some opinions on this though. While it does work very well, the theming itself would then rely on providing these rgb values rather than providing the values of the variables themselves. I tried looking at other options for styling the box shadows but they didn't work; it was either something like this that kept it dynamic and tied together with the non-box-shadow values, or creating an entirely new variable that was only for box shadows that was also staticly set like:
```scss
:root {
  --focus-color: $blue-100;
  --focus-color-shadow: rgba($blue-100, $alpha-600)
}
```
Do you guys thing this makes it needlessly complicated? or does it make sense?
Theming for variables that have rgb values that are used for other things like box-shadows would have to be in the following format: `--brand-color-primary-rgb: 133, 59, 148;` and that would propagate to everything. And for values that do not have other elements tied to it that change the alpha channel you would still be able to theme the variable itself like so: `--brand-color-primary: rgb(133, 59, 148);` but in the spirit of consistency I added the the mixin to all of the themable colors just so you don't have multiple apis.

ewdicus: Instead of using a mixin to get two variables, could we provide an alternative function to `rgba(var(--var-name), alpha);`?

Something like (warning: not super familiar with scss functions)
```scss
@function rgba-var($color-value, $alpha-value) {
  @return "rgba(" + red($color-value), green($color-value), blue($color-value) + ", " + $alpha-value + ")";
}
```
And then use that like:
```scss
//...
box-shadow: 0 0 modular-scale-px(-6) 9
      rgba-var(--step-progress-active-color, $alpha-600);
//...
```

ahmedmodan: My mistake the static variable should actually be something like this:

```scss
:root {
  --focus-color: $blue-100;
  --focus-color-shadow: rgba(127, 234, 255, $alpha-600)
}
```

ewdicus: Ah, I think I answered my own question. I'm mixing up what's happening at scss compile time and what we're substituting later.

ahmedmodan: yeah the `rgba` function is both a scss function and a css function. The scss function will accept hex, hsl, and rgb values, while the css function will only accept rgb values

ahmedmodan: the `red`, `green`, and `blue` functions are also only scss. I used them here since i didn't want to convert all of the existing hex color variables to rgb.

gelsto: Actually now that I'm looking at how it's used I'm not so bothered by the extra complexity. I guess you'll just end up with a bunch of generated variables that you don't use, which doesn't seem too bad right now?

ahmedmodan: Yeah we will have extra variables that are unused. The upside that I see in this is that we don't have to worry about changing a variable like `--focus-color` and then enforcing the change of `--focus-color-shadow`, either via code-review and making sure we catch it, or when we create the admin tool version of this having custom logic for dependent rules. The downside is that there are more variables, and likely more documentation and using the both the normal variables and the rgb variables.

gelsto: Yeah I think something like `--x-color-rgb-values` or `--x-color-rgb-parameters` would be a better variable name here. This might be a silly question but is there a performance issue with more variables?

ahmedmodan: Yeah I like `--x-color-rgb-values`. I don't think there is a noticable performance hit with variables unless you are actively changing them. Since changing them will require a layout re-calculation. Since we only really set them once and then forget about it, it shouldnt be much of an issue.

sumul: @ahmedmodan I think this is a good approach. It makes it pretty obvious what to do when you're setting theme colors for a new partner. One thing that's less obvious is how to enforce that the new mixins are used when implementing a new component. To that end, we will need to write some documentation about how and when to create new themeable variables when building new components in Phenotypes, and also how to best use the existing variables.

Speaking of documentation, I think this PR looks great but should probably include an update to the documentation about text color utility classes to show off the new ones you added.

ahmedmodan: Yeah for sure. I'll update the variable name to `--x-color-rgb-values`. And also add more docs for theming and contributing.

guides/11-theming.md
```diff
@@ -0,0 +1,146 @@
...
+
+**NOTE:**
+For a good chunk of variables there are two variable declarations, with values from one relying on the other. For example the `--focus-color` variable relies on the `--focus-color-rgb-values` variable. This is done to allow styling of elements that are tied to each other color wise but don't necessarily need to be defined twice. Phenotypes allows you to customize focus color, but that also means that the box-shadow that provides the focus color "glow", must also match that customized focus color. In order to ensure that you don't need to define two different variables (`--focus-color` & `--focus-shadow-color`) we expose a top level variable that defines the rgb values that should be used for both. So in this case changing `--focus-color-rgb-values` will apply the change to both the focus color as well as the box-shadow that is tied to that focus color. **It is for this reason that we recommend setting the rgb-values and allowing those values to update variables downstream.**
+
+```css
```
gelsto, line 19: I like that this is more informative than just linking to `_theme_config.scss`, but we might want to add a comment in that file bugging us to update this documentation whenever we make changes in there.

README.md
```diff
@@ -106,6 +107,8 @@ $grid-breakpoints: (
 
 The `_mixins.scss` file contains imports for functions and mixins that are used by other modules and components. Actual implementations should go in `styles/mixins`. E.g. `styles/mixins/_type.scss` contains both functions and mixins that help with typography.
 
+The `_theme_config.scss` file contains a css ruleset that defines the theming css variables that phenotypes uses. These are defined as css custom properties and allow custom themes for Phenotypes. Please see the [Theming guide](/guides/11-theming.md) for more info
```
gelsto, line 110: missing '.'

README.md
```diff
@@ -145,6 +148,18 @@ The following command will install Phenotypes into `@aminohealth/phenotypes`:
...
 ```
 
+### SASS
+
+The recommended way to use the styles for phenotypes is to import the scss entrypoint phenotypes.scss in your own project. When installed via npm this will be located at `@aminohealth/phenotypes/styles/phenotypes.scss`. This does require your project to use [Sass](https://sass-lang.com/). This is; however, not required and for that reason we also provide plain old css that you are able to use.
+
+**NOTE**  By default the phenotypes.scss entry point when compiled will include CSS variables in the output. CSS variables are required for custom themes. CSS variables will work fine for all evergreen browsers; however, if you need to support a browser that does not support CSS variables (IE) or do not require a custom theme, you can compile the css variables away (Or see the [Theming guide](/guides/11-theming.md) for supporting IE .) CSS variables can be compiled away easily by post-processing via [postcss](https://postcss.org/) and using the [postcss-preset-env](https://preset-env.cssdb.org/) plugin. An example of how to do this is can be seen in our own build process when we compile the phenotypes.scss file to plain old css [here](./scripts/compileScss.js#L40-L48).
```
gelsto, line 155: "(Or see the [Theming guide](/guides/11-theming.md) for supporting IE .)" -> "(or see the [Theming guide](/guides/11-theming.md) for supporting IE)."

guides/11-theming.md
```diff
@@ -0,0 +1,146 @@
+---
+title: Theming
+status: draft
+---
+
+Phenotypes allows theming for certain aspects of the library. This is done via css custom properties that can be overridden by you as the consumer in your own application.
+
+## Browser Support
```
gelsto, line 8: Does this section belong in the main README?

ahmedmodan: I'll add it to the main readme also. but I think it is also helpful to add to the theming guide on the fractal page

PR #113, opened 2020-09-29
Update grid for large breakpoint

ewdicus:
In new designs we're now including a "large" version to use more of the available screen real estate. This PR updates the max-width of the container and the gutter size for the grid at the large breakpoint.

### DIFF
--- a/fractal_assets/css/phenotypes.themable.css
+++ b/fractal_assets/css/phenotypes.themable.css
@@ -2208,12 +2208,12 @@ hr {
       padding-right: 53px; } }
   @media (min-width: 1200px) {
     .container {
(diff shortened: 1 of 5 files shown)


ahmedmodan (changes requested):
@ewdicus can you also run the app locally to propagate those changes to the compiled css? I think you should be able to do `make build dev`. we check in changes to the compiled css stylesheets. It will also be run as a `prepublishOnly` when you attempt to publish to npm. Ideally we probably shouldn't check in the compiled css but we can leave that to a breaking change in the future.

ahmedmodan (approved):
tenor-65106135

PR #114, opened 2020-10-09
Phenotypes colors refresh

sumul:
This is the first major change to a core aspect of Phenotypes: our color palette. It's something I've been working on for a while. I'm excited to share it.

The overall goals are to:

* Simplify the color system by removing colors we don't use and are unlikely to use
* Make the color system more vibrant and attractive for both product and marketing purposes
* Make interactive elements like buttons and links feel more interactive, in particular by nudging the blue hues in the direction of the blues used in OSes and browser defaults
* Get rid of some quirks and hacks (e.g. gradients using nonstandard values, core product headers using nonstandard values)
* Stay true to the existing brand look and feel overall. For most people, this should feel like a non-event and old Amino assets should still feel roughly "on-brand."

Here are the old colors: 

old colors

And the new:

colors


styles/_variables.scss
```diff
@@ -15,39 +15,42 @@
 // ------------------------------
 // Colors
 // ------------------------------
+// Note: some of the original -400 and -500 colors were deprecated
+// as part of the colors refresh in October 2020. They are mapped
+// to the next available color for backwards compatibility.
```
NickDunkman, line 20: 👍

NickDunkman (approved):
Love it!

PR #119, opened 2021-02-16
A11y Colors

sumul:
Updates all base color variables and themeable color variables to meet WCAG 2.1 contrast ratio guidelines.

image


ahmedmodan (approved):
Looks great! going to make this a major version bump since colors have changed.

PR #124, opened 2021-04-13
audit and upgrade packages

orjames:
pivotal: https://www.pivotaltracker.com/story/show/177744167

This audits all the security vulnerabilities in phenotypes and upgrades the packages accordingly.

From: 
```
1384 vulnerabilities found - Packages audited: 1330
Severity: 1129 Low | 12 Moderate | 242 High | 1 Critical
```

To:
```
2 vulnerabilities found - Packages audited: 1169
Severity: 2 Low
```

The remaining 2 vulnerabilities are from the `@aminohealth/fractal-react-adapter` package, which will be audited as well

with the updates this builds and runs, and was tested locally

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
     "classnames": "^2.2.5",
     "keycode": "^2.1.9",
     "node-sass": "^4.14.1",
-    "react": "15.4.2",
-    "react-dom": "15.4.2"
+    "react": "17.0.2",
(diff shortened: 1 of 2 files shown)


ahmedmodan (approved):
LGTM. We will need to create a release. I think it makes sense to make this a major release since we are upgrading to react 17 from react 15 here.

PR #127, opened 2021-05-06
add aria to associate error with form

orjames:
Fix accessibility issues related to error messages
```
The error messages on forms show visibly but are not programmatically associated with the text field potentially causing users to not understand how to complete the forms. 
```

this shows how the `aria-errormessage` is getting the proper id:
Screen Shot 2021-05-06 at 2 35 32 PM

screen reader alerting "required" once the `save` button is clicked:
Screen Recording 2021-05-07 at 10 28 11 AM

more info on `aria-errormessage: [https://www.w3.org/TR/wai-aria-1.1/#aria-errormessage](https://www.w3.org/TR/wai-aria-1.1/#aria-errormessage)

### DIFF
--- a/library/components/form_group/form_group.jsx
+++ b/library/components/form_group/form_group.jsx
@@ -27,7 +27,7 @@ class FormGroup extends React.Component {
     return (
       errorBoolean && (
         <div
-          id="formErrorMessage"
+          id={this.props.controlId}
           className="FormGroup__error"
           aria-live={errorBoolean ? 'assertive' : 'off'} // A live region scenario is when an error message is displayed to users only after they have provided invalid information
         >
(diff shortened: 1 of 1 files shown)


library/components/form_group/form_group.jsx
```diff
@@ -23,9 +23,16 @@ class FormGroup extends React.Component {
...
   renderError() {
+    const errorBoolean = !!this.props.error;
     return (
-      !!this.props.error && (
-        <div className="FormGroup__error">{this.props.error}</div>
+      errorBoolean && (
+        <div
+          id={this.props.errorId}
+          className="FormGroup__error"
+          aria-live={errorBoolean ? 'assertive' : 'off'} // A live region scenario is when an error message is displayed to users only after they have provided invalid information
```
ewdicus, line 32: Since this is only rendered when errorBoolean true, I don't think it's possible for this to ever be "off".

library/components/form_group/form_group.jsx
```diff
@@ -23,9 +23,16 @@ class FormGroup extends React.Component {
   }
 
   renderError() {
+    const errorBoolean = !!this.props.error;
```
ewdicus, line 26: This might be a little easier to follow as something like `hasError`, so it's obvious that the "true" state means that there is an error.

ewdicus (changes requested):
💪🏻 I think this seems reasonable, some small suggestions.

ewdicus (approved):
Cool, thanks for making those changes

PR #137, opened 2021-06-16
Border sizing fix for buttons

orjames:
Updates the primary color and error color buttons to have a 1px border of the same color, to keep the height the same as the already bordered secondary buttons


ewdicus (approved):
Just a note that Owen and I discussed the alternative of specifying the height (which, in conjunction with border-box, would also fix this), but this felt like the more flexible option.

PR #143, opened 2021-06-25
Build static site

ewdicus:
This PR adds the setup necessary to build the phenotypes site statically.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ jspm_packages
 .node_repl_history
 
 dist/
+static-build/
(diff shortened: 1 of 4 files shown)


ahmedmodan (commented):
The setup and commands look good. I can see the static site is built and view it locally.  but I can't see the components in the components pages. Looks like some cross origin error with mandelbrot. Do you run into the same issue?

ewdicus:
@ahmedmodan I get something similar when opening index directly in a browser. Firefox at least handles things a bit differently when it's loading something from file://. If you shell into the static-build directory and run `python3 -m http.server` that will make the files available on localhost:8000.

ahmedmodan (approved):

PR #144, opened 2021-07-12
Replace deprecated node-sass with sass

NickDunkman:
`node-sass` is deprecated and does not run on the new Apple silicon! The node-sass people [say to upgrade to "dart" sass](https://sass-lang.com/blog/libsass-is-deprecated), which is just called `sass`.

With this change, phenotypes can now be built & run on the new m1 chips.

A couple trivial changes:

- the `/` operator is deprecated in dart sass -- had to convert them all to use `math.div`
- there are a bunch of changes in the compiled css files, but they're all just replacing single quotes with double quotes, and whitespace changes.

### DIFF
--- a/configs/rollup.fractal-development.config.js
+++ b/configs/rollup.fractal-development.config.js
@@ -35,6 +35,6 @@ export default {
     }),
   ],
   // we don't want to watch and rebuild when the scss files change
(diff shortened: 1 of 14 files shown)


gelsto (approved):
looks good! thanks for updating 🎉

PR #148, opened 2021-07-15
Simple nginx config

jimfingal:
@ewdicus cribbing off of how we do it in the web, here's a proof of concept for putting nginx in front of fractal. It involves:

* Building the static site as part of the base docker build
* Mapping volumes between the fractal container + the nginx container
* Symlinking the static output folder to be where nginx by default looks for static html files

With this you should be able to `make build run` + visit http://localhost:3001 to see how nginx renders the static site.

Am I missing anything that would pose issues to this approach?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,14 +2,22 @@ FROM node:14.4.0-stretch-slim
 
 ENV NPM_CONFIG_LOGLEVEL warn
 
+ENV APP /usr/src/app
+
 # Create app directory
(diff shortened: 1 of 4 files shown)


ewdicus:
At first look, this seems great. I'll pull this locally and poke at it a bit as well

jimfingal:
Going to add some updates to amino_build, and fix the volume mapping for docker-compose 3...

PR #153, opened 2021-09-03
fix button--link class

orjames:
Update the Button--link class in phenotypes to more accurately portray a button styled as a link.

The old:
Image 2021-09-03 at 3 13 07 PM

The new:
Image 2021-09-03 at 3 12 35 PM


NickDunkman:
@orjames Are we no longer relying on the `.Button--link` as it was before? This is a replacement for the `.LinkButton` BEM module in pare_web?

orjames:
> @orjames Are we no longer relying on the `.Button--link` as it was before? This is a replacement for the `.LinkButton` BEM module in pare_web?

we actually don't use Button--link anywhere in Pare_web (my guess is because it's broken) this change came about because here: https://github.com/parelabs/pare_web/pull/4608, we can't just pass in the BEM `.LinkButton` class but need to change the phenotype button component.

PR #154, opened 2021-09-13
Encode checkbox indicator background URL replacing spaces with %20

jessiditocco:
**Linear Story**:
https://www.pivotaltracker.com/story/show/179258172

**Overview**
- **Reported bug**: It was discovered that all phenotype checkboxes are missing their checkbox.
- **Cause of bug**: Recently, we switched to rcss to minify our CSS ([PR](https://github.com/parelabs/pare_web/pull/4494)). This minifier expects spaces to be encoded, which is why the checkboxes aren't showing as expected because the checkbox in Phenotypes uses an svg as a background image.
- **Bug fix**: This PR updates the svg URL used by the `checkbox-icon` background image by properly encoding the spaces in the URL


library/components/checkbox/_checkbox.scss
```diff
@@ -50,7 +50,7 @@
 
 @mixin checkbox-icon($fill: $white, $fill-opacity: 1) {
   background-image: str-replace(
-    url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 10 9'%3E%3Cpolygon fill='#{$fill}' fill-opacity='#{$fill-opacity}' points='3.5 5.45 1.5 3.45 0 4.95 3.5 8.45 10 1.95 8.5 .45'/%3E%3C/svg%3E%0A"),
+    url("data:image/svg+xml;charset=utf8,%3Csvg%20xmlns='http://www.w3.org/2000/svg'%20viewBox='0%200%2010%209'%3E%3Cpolygon%20fill='#{$fill}'%20fill-opacity='#{$fill-opacity}'%20points='3.5%205.45%201.5%203.45%200%204.95%203.5%208.45%2010%201.95%208.5%20.45'/%3E%3C/svg%3E%0A"),
```
gelsto, line 53: We should leave a note about why this needs to be encoded. In the rss issue, they point to [rfc2397](https://datatracker.ietf.org/doc/html/rfc2397) as why they remove spaces from these urls.

We can also use variables to make this a little more future-proof/easier to read:
```scss
@mixin checkbox-icon($fill: $white, $fill-opacity: 1) {
  $svg-url: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 10 9'%3E%3Cpolygon fill='#{$fill}' fill-opacity='#{$fill-opacity}' points='3.5 5.45 1.5 3.45 0 4.95 3.5 8.45 10 1.95 8.5 .45'/%3E%3C/svg%3E%0A");
  $svg-url: str-replace($svg-url, "#", "%23");
  $svg-url: str-replace($svg-url, " ", "%20");

  background-image: $svg-url;
}
```

jessiditocco: Added a comment pointing to rfc2397!

Also using the str-replace function for both the `#` and spaces is way cleaner 🧼

gelsto (approved):
🎉 💪 Thanks for fixing! This was a tricky one

PR #161, opened 2021-11-22
Add an "*" to labels of required fields, for accessibility 

NickDunkman:
One of the items in our accessibility audit was making required fields more obviously required. My instinct was to append an asterisk to the labels of required fields, as its very common, unlikely to disrupt our layouts, and appears to be an accepted way of doing this: https://www.w3.org/WAI/WCAG21/Techniques/html/H90 (see example 2).

Also, looks like our js tests no longer work, since Enzyme isn't compatible with React 17. To fix, I upgraded Enzyme & also pulled in an unofficial adapter that everyone seems to be using until the Enzyme peeps release their own *([@wojtekmaj/enzyme-adapter-react-17](https://www.npmjs.com/package/@wojtekmaj/enzyme-adapter-react-17)). It's a dev dependency, so shouldn't be an issue.


library/components/form_group/_form_group.scss
```diff
@@ -60,12 +56,26 @@
...
+    @include form-group-sizer("small");
+  }
+  .FormGroup--large {
+    @include form-group-sizer("large");
+  }
+}
+
+abbr[title].FormGroup__required {
+  text-decoration: none;
 }
```
NickDunkman, line 81: Added this to remove the automatic dotted underline that modern browsers add to `abbr[title]` elements -- felt like in this case where it's just an asterisk at the end of a label, it feels gross:

Screen Shot 2021-11-22 at 10 36 11

The rule is pretty specific, but was necessary to override a style in our reset file.

(the rest of these changes are all prettier formatting)

ahmedmodan (approved):
tenor-157944541

PR #170, opened 2022-01-19
Add end-of-file-new-line settings

NickDunkman:
Using the same settings we use in our other Amino repos -- I ran through all of the files I had touched and re-formatted/re-saved. Everything shows a new line in VS Code, but there were only 2 files affected in the diff 🤷


amino_build.json
```diff
@@ -12,4 +12,3 @@
         }
     }
 }
-
```
NickDunkman, line 15: this one had TWO new lines at the end

ahmedmodan: 😱

ahmedmodan (approved):

PR #174, opened 2022-01-20
convert display utilities to storybook format

NickDunkman:
POC of "utility" doc. Heavily lifting things from [Bootstrap's docs](https://getbootstrap.com/docs/5.0/utilities/display/) -- their licensing allows it tho! I updated our attribution text to fully abide by the terms of the license.

Screen Shot 2022-01-26 at 13 45 35


src/utilities/display/display.stories.mdx
```diff
@@ -0,0 +1,86 @@
...
+[display property](https://developer.mozilla.org/en-US/docs/Web/CSS/display) of
+an element.
+
+### Notation
+
+Display utility classes that apply to all
+[breakpoints](?path=/docs/guides-layout--page#responsive-breakpoints), from `xs`
+to `xl`, have no breakpoint abbreviations in them. This is because those classes
+are applied from `min-width: 0;` and up, and thus are not bound by a media
+query. The remaining breakpoints, however, do include a breakpoint abbreviation.
```
ahmedmodan, line 17: This description immediately followed by the example seems confusing. I know the explanation that these utilities apply to the breakpoints and up is written below on line 35, but upon initially reading this description,  i come to the conclusion that no breakpoint means it applies to everything, but the examples given on lines 21-22 seems to then indicate otherwise (with the `'.d-value' for 'xs'` phrasing). What do you think of moving the explanation on lines 35-38 up to be right after this description. Or perhaps rephrasing the example notation on lines 21-22 to be more clear  that each of those examples apply to the breakpoint and up?

ahmedmodan (approved):
looks good. one suggestion or comment on the order of the descriptions for describing how utility classes are applied to what breakpoints.

src/utilities/display/display.stories.mdx
NickDunkman: Good point -- it is a bit confusing as is! I'll stew on it for a bit (probably want to develop some language that will be used in any breakpoint-modifier docs)

NickDunkman: @ahmedmodan I did a little bit more of a re-imagining -- what do you think of this?

Screen Shot 2022-01-27 at 15 36 44

NickDunkman: slight tweak:

Screen Shot 2022-01-27 at 15 39 33

ahmedmodan (approved):

PR #172, opened 2022-01-19
Convert Radio to typescript & storybook format

NickDunkman:
Presenting this `Radio` button conversion as a POC for our component doc style.

* There is a nice `mdx` file shown under the "Docs" tab that explains everything in sensible groupings.
* The `mdx` file contains `<Story>` definitions that also magically appear on the sidebar -- you can toy with the props for each in the "Canvas" tab.
* Storybook does not by default include all the native `<input>` props, which are included in the `Radio` component props. Rightly so, as there are hundreds of them, and they'd make the storybook docs very noisy. I decided to re-define the types of the more commonly used props in our own `RadioProps` type, so that they can appear more quickly in vscode (without having to recursively click through several prop definitions, have better doc strings, and so that they appear in the props table at the top of the docs.

Screen Shot 2022-01-19 at 15 06 10

PR #178, opened 2022-01-28
GUIDE-479: convert TextInput component to typescript + story format

NickDunkman:
Followed the same patterns I used in the [Radio conversion](https://github.com/aminohealth/phenotypes/pull/172)!

ezgif-5-6e0633be79


gelsto (approved):
This lgtm! I see that we hammered out this pattern nicely in the previous PR, so that context is helpful. I'm curious about if we're planning on using cavases to detail most props like this going forward, or only for a subset of components, vs. a single story where the user can customize the props to see different combinations.

NickDunkman:
> I'm curious about if we're planning on using cavases to detail most props like this going forward, or only for a subset of components, vs. a single story where the user can customize the props to see different combinations.

@gelsto I was thinking of using this general pattern for all components -- I think it's nice to see the full capability of the component (to an extent) in the docs -- you can still click into individual stories (and the "canvas" tab) to customize props. What are your thoughts?

gelsto:
I think the tradeoff is auto-updating vs. more detailed/complete documentation. This is kind of a weird example because a text input is very simple, but if you had props that are likely to be tweaked over time, you could rely on the story to reflect the full capability of the component without having to update the story at all.

NickDunkman:
@gelsto Cool -- yea I think that the stuff we have in phenotypes is generally a good candidate for more thorough docs, the components we have here are more stable than usual.

PR #179, opened 2022-02-02
GUIDE-480: Convert Button component to typescript + story format

NickDunkman:
Got the typescript working! E.g.:

- You get yelled at if you pass an `href` prop and then try to use `button`-specific props
- You get yelled at if you pass `a`-specific props but don't include an `href`
- If you're doing things correctly you don't get yelled at 😊 

Screen Shot 2022-02-02 at 14 15 45


fractal_assets/css/phenotypes.themable.css
```diff
@@ -595,15 +595,15 @@ hr {
   background: var(--primary-button-active-color);
 }
 
-.Button--link .Button__control .Button__control {
+.Button--link .Button__control {
```
NickDunkman, line 598: (we had a bug in our `.Button--link` styles)

package.json
```diff
@@ -65,7 +65,7 @@
     "buildCss": "node scripts/buildScss.js",
     "watchCss": "npm run buildCss && node scripts/watchScss.js",
     "startProd": "npm run buildCss && npm run storybook",
-    "start": "npm run storybook && npm run watchCss",
+    "start": "npm run buildCss && npm run storybook",
```
NickDunkman, line 68: this is temporarily better for me in development -- have to restart storybook to pick up css changes anyway. Will be simplified to just the following once we get sass hot-reloading working:

```
"start": "npm run storybook",
```

src/components/Button/AnchorButton.tsx
```diff
@@ -0,0 +1,96 @@
...
+import classes from 'classnames';
+import keycode from 'keycode';
+import { AnchorButtonProps } from './types';
+
+/**
+ * A Button rendered using an anchor (`<a>`) element, for buttons that link to
+ * an href.
+ * @param props {@link AnchorButtonProps}
+ */
+export function AnchorButton({
```
NickDunkman, line 11: All of the logic in this component is the same as before -- I did refactor from a class component to function component, though!

src/components/Radio/Radio.tsx
```diff
@@ -3,7 +3,7 @@ import classes from 'classnames';
 
 interface RadioProps extends React.ComponentProps<'input'> {
   /** Adds a label to the radio */
-  children?: React.ReactChild;
+  children?: React.ReactNode;
```
NickDunkman, line 6: (realized this was the better way to type children when working on the Button)

src/components/Button/AnchorButton.tsx
```diff
@@ -0,0 +1,96 @@
...
+  disabled = false,
+  href,
+  onClick,
+  tabIndex,
+  ...anchorHTMLProps
+}: AnchorButtonProps) {
+  const anchorRef = React.useRef<HTMLAnchorElement>(null);
+  const [spaceIsPressed, setSpaceIsPressed] = React.useState<boolean>(false);
+
+  const handleClick = React.useCallback(
```
ahmedmodan, line 22: I don't think any of these should be in a use callback. `useCallback` is only useful if you are passing the callback to an optimized child component that you don't want to re-render frequently. In this case the child component is just going to be the DOM element and the `useCallback` is just unnecessary overhead.

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+
+import { AnchorButton } from './AnchorButton';
+import { TrueButton } from './TrueButton';
+
+type ButtonProps =
+  | React.ComponentProps<typeof TrueButton>
+  | React.ComponentProps<typeof AnchorButton>;
```
ahmedmodan, line 8: why ComponentProps? The props for `TrueButton` and `AnchorButton` are already defined in a separate types file and exported from that file. We can import and use them directly.

ahmedmodan (commented):
Would definitely recommend getting rid of `useCallback` since it gives us no benefit here.

src/components/Button/Button.tsx
NickDunkman: 👍 Wrote this before I separate those types out into types.ts -- good catch.

https://github.com/aminohealth/phenotypes/pull/179/commits/b7c2ccbfb1f64248fd2141e3f2956cd6f4dc0257

src/components/Button/AnchorButton.tsx
NickDunkman: 👍 https://github.com/aminohealth/phenotypes/pull/179/commits/f6cfa5fa56c853464801294fb04259574ac7068a

ahmedmodan (approved):

PR #181, opened 2022-02-03
Guide-482: convert FormGroup to typescript & story format

NickDunkman:
I didn't bother with reproducing _all_ of the accessibility detail from the previous README in these docs, since we're going to remove `FormGroup` from the public API soon.

Screen Shot 2022-02-03 at 15 44 25


gelsto (approved):
Looks good! Agreed that we don't need to port all the documentation if this is going to be short lived 👍

PR #187, opened 2022-02-10
Change active Tab bottom-border to always match the text color

NickDunkman:
This makes the bottom border automatically themeable (as long as it is set to a themable `color`, which active Tabs are, by default).

Also enables the use case in our new care dashboard where we have white tabs (white text + white lines) on the purple background -- you can just do this:

```jsx
<Tab active className="text-color-reversed-primary">Care</Tab>
```

After this change --- default "black" (is actually 86% black now) border still works:

Screen Shot 2022-02-10 at 14 50 33

Setting color to red like this:

```
<Tab active style={{ color: red; }}>Normal Tab</Tab>
```

looks like:
Screen Shot 2022-02-10 at 14 54 39


library/components/tab/_tab.scss
```diff
@@ -9,5 +9,10 @@
 }
 
 .Tab--is-active {
-  border-bottom: 2px solid $black;
+  // The color property is ommitted so that the border is the same color as
+  // the Tab's text (the `color` property). That makes it easy to have
+  // "reversed" Tabs -- you can have a tab with white text & a white underline
+  // like this:
+  //   <Tab className="text-color-reversed-primary">Settings</Tab>
```
ewdicus, line 16: 👏🏻

ewdicus (approved):

PR #191, opened 2022-02-15
GUIDE-484: convert Slider to typescript + storybook

NickDunkman:
Ended up abandoning the effort to make the Slider compatible with Formik. That'd be easier in the future as a full rewrite (or build on top of some other already compatible slider library).

Screen Shot 2022-02-15 at 15 26 04


src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,295 @@
...
+ * [controlled component](https://reactjs.org/docs/forms.html#controlled-components).
+ * @param props {@link SliderProps}
+ * @returns
+ */
+export function Slider({
+  className,
+  disabled = false,
+  max = 100,
+  min = 0,
+  step = 1,
```
gelsto, line 100: Do these make sense as defaults? I wonder if they should just be required props. (I see this is just ported over, so not blocking)

gelsto (approved):
Looks great! Left one nb thing

src/components/Slider/Slider.tsx
NickDunkman: Yea good call out! That was my thought too -- but the MDN docs say that these are the defaults used by the native `<input type="range" />` -- so seems best practicey?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#validation

I could also see these defaults _never_ being used.

gelsto: Ah, those being the defaults for `range` seems like a reasonable argument. Agree that they probably won't be omitted very often, even in situations where we're using 1-100 anyway

NickDunkman: Yea. I say let's leave them -- it's a nice extra hint about how to use the component when you hover over the component in VSCode

PR #192, opened 2022-02-18
GUIDE-502: convert RadioGroup to typescript + storybook

NickDunkman:
I also made RadioGroup compatible with formik! It should be useful in pare_web now -- though we'll still need to wrap it with legend/hint/error stuff until we work on phenotypes form molecules later on ...

This is working code now!

```tsx
<formik.Formik initialValues={{ drink: 'coffee' }}>
  <formik.Form>
    <formik.Field
      as={phenotypes.RadioGroup}
      name="drink"
      options={[
        { value: 'coffee', label: 'Coffee' },
        { value: 'tea', label: 'Tea' },
      ]}
    />
  </formik.Form>
</formik.Formik>
```

Screen Shot 2022-02-17 at 17 52 36

Screen Shot 2022-02-17 at 17 56 43


src/helpers/errors.ts
```diff
@@ -0,0 +1,21 @@
...
+  printError(
+    `Warning: ${componentName} called with both ${controlledPropName} and ${uncontrolledPropName} props. ${componentName} must be either controlled or uncontrolled (specify either the ${controlledPropName} prop, or the ${uncontrolledPropName} prop, but not both). Decide between using a controlled or uncontrolled input element and remove one of these props. More info: https://reactjs.org/link/controlled-components`
+  );
+}
+
+export function printValueWithoutOnChangeError(componentName: string) {
+  printError(
+    `Warning: You provided a \`value\` prop to ${componentName} without an \`onChange\` handler. This will render a read-only field. If the field should be mutable use \`defaultValue\`. Otherwise, set either \`onChange\` or \`readOnly\`.`
+  );
+}
```
NickDunkman, line 21: This is mimic-ing what React does when you use native form elements in these improper ways -- they log an error to the console with basically these exact messages.

src/components/RadioGroup/_radio_group.scss
```diff
@@ -0,0 +1,15 @@
...
+
+    &.Radio--large {
+      margin-top: $spacer-5;
+    }
+  }
+}
+
+.RadioGroup__Radio--inline:not(:first-child) {
+  margin-left: $spacer-4;
+}
```
NickDunkman, line 15: brought these in from pare_web

src/components/Checkbox/Checkbox.tsx
```diff
@@ -7,7 +7,7 @@ interface CheckboxProps extends React.ComponentProps<'input'> {
    */
   className?: string;
   /** Sets the Checkbox to checked when using as a
-   * [controlled component](https://reactjs.org/docs/forms.html#controlled-components) */
+   * [controlled component](https://reactjs.org/link/controlled-components) */
```
NickDunkman, line 10: This URL 301s to the one I replaced it with -- I noticed that React's error messages use this URL, so it must be the more external URL they intend to maintain for this particular documentation.

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,160 @@
+import * as React from 'react';
+import classes from 'classnames';
+
+import {
+  printControlledAndUncontrolledError,
+  printValueWithoutOnChangeError,
+} from '../../helpers/errors';
+import { Radio } from '../Radio';
+
```
ahmedmodan, line 9: What do you think of aliasing `React.ComponentProps<'input'>`? May be easier to read. But I am also ok leaving as is.

```ts
type ReactInputProps = React.ComponentProps<'input'>
```

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,160 @@
...
+    !isControlledComponent && setUncontrolledValue(event.target.value);
+    onChange?.(event);
+  }
+
+  function handleRadioBlur(event: React.FocusEvent<HTMLInputElement>) {
+    onBlur?.(event);
+  }
+
+  // This function will generate Props to pass to each Radio
+  function managedRadioPropsGenerator({
```
ahmedmodan, line 108: I would recommend renaming this. maybe `getManagedRadioProps` or just `radioProps` like you have in the storybook example? To me when I read `Generator` I assumed that we were somehow using a [generator function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*) which could potentially be confusing.

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,160 @@
...
+    // When there is an `options` prop, apply some standard layout logic. (If
+    // there is a render function instead, it is expected that the parent is
+    // taking full control of the layout.)
+    const layoutProps = options
+      ? {
+          className: inline
+            ? 'RadioGroup__Radio--inline'
+            : 'RadioGroup__Radio--block',
+        }
+      : null;
```
ahmedmodan, line 124: This can be moved to a normal prop for the `Radio` component on line 144/145. Since this is only relevant when there is an options prop, we can inline it on the prop itself rather than having a ternary in this function

src/components/RadioGroup/RadioGroup.stories.mdx
```diff
@@ -0,0 +1,157 @@
...
+  Story,
+  Description,
+  Source,
+} from '@storybook/addon-docs';
+import { dedent } from 'ts-dedent';
+
+import { RadioGroup } from './RadioGroup';
+import { Radio } from '../Radio';
+
+export const Template = (args) => <RadioGroup {...args} />;
```
ahmedmodan, line 14: Can we add fieldset and legend to these examples? The radio group shouldn't be used without those for accessibility reasons and the examples should reflect that until we make those part of the radio group itself and the consumer doesn't need to wrap manually.

guides/12-formik.stories.mdx
```diff
@@ -77,6 +77,23 @@ implementation hints.
...
+      <formik.Field
+        as={phenotypes.RadioGroup}
+        name="drink"
+        options={[
+          { value: 'coffee', label: 'Coffee' },
+          { value: 'tea', label: 'Tea' },
+        ]}
+      />
+    </formik.Form>
+  </formik.Formik>
```
ahmedmodan, line 94: add fieldset and legends to these also

ahmedmodan (commented):
Would recommend updating the usage examples to account for proper usage to adhere to a11y guidelines until we can make `fieldset` and `legend` part of the component itself. Also would recommend updating the name of the render callback.

src/components/RadioGroup/RadioGroup.tsx
NickDunkman: I find the `<'input'>` to make it _super_ obvious that it is pulling directly from the native `input` element 🤷

src/components/RadioGroup/RadioGroup.tsx
NickDunkman: Smart! https://github.com/aminohealth/phenotypes/pull/192/commits/946555db17ee772b8cb00e1147a51587a1831d74

NickDunkman:
As discussed offline with @ahmedmodan -- given that a11y _requires_ the usage of a `<fieldset>` with a group of radios, decided to go ahead and build that into RadioGroup instead of just adding as a wrapper in the docs examples.

Since the legend has to be inside the fieldset, I added legend+hint as well, so that this covers our existing pare_web use cases. The legend is technically not required by a11y, so left that optional for now, but we'll probably want to make it required when we work on form field molecules later.

Added this section to the docs:

Screen Shot 2022-02-18 at 14 58 03

PR #193, opened 2022-02-18
GUIDE-488: convert Message docs to storybook + add new React component

NickDunkman:
We currently do not have a react component for this -- pare_web just hand crafts a `div` with the proper classNames. Decided to go ahead and add a React component, even though it's very simple, to make it consistent with all of our other phenotypes components.

Eventually we will add modifier props for all our components -- so you can do something like this and get typescript errors if you use a bad modifier:

```jsx
<Message danger>OH NO</Message>
```

However, going to defer on that and work on modifier props later for our full component library, so for now keeping things consistent with all our other components:

```jsx
<Message className="Message--danger">OH NO</Message>
```

Screen Shot 2022-02-18 at 12 14 16


src/components/Message/Message.tsx
```diff
@@ -0,0 +1,20 @@
+import classes from 'classnames';
+import * as React from 'react';
+
+interface MessageProps {
+  /** The contents of the Message */
+  children?: React.ReactNode;
+  /** Three styles are available via this className pattern:
+   * `Message--{danger|info|success}`
+   */
+  className?: string;
```
jessiditocco, line 10: Should this type also include "null"?
```suggestion
  className?: string | null;
```

I've seen this pattern in other places in our code base where we have 
```
  className?: string | null | undefined;
```

Logan explained that we actually don't need the undefined because having the `?` after classname implicitly adds that to the type of className, but do we still need null?

jessiditocco (approved):
Looking good! One typescript question I'm unsure about but other than that this looks good to merge 🚢

src/components/Message/Message.tsx
NickDunkman: Good question! We do have some legacy code in pare_web that is include the `null` type -- but moving forward I think it's better to match how the native dom elements type their `className` prop as just `string | undefined`:

Screen Shot 2022-02-18 at 13 15 47

So the proper way to type it is:

```ts
propName?: string;
```

PR #195, opened 2022-02-25
GUIDE-511: convert typography utilities to storybook

NickDunkman:
I really like how Bootstrap documents their [css utilities](https://getbootstrap.com/docs/5.1/utilities/), and we have a lot of overlap with them anyway, so I broke our typography utilities into 3 separate systems/docs like they do: `text`, `overflow`, and `color`. I also conservatively pulled in a bunch of utilities for those 3 groupings from pare_web (I was rewriting all of the docs, and leaving them omitted felt upsetting). All of these changes are backwards compatible (except for a few things that no longer make sense, like some old colors). There are no changes to the theme variables or their values.

I built a couple of utility docs "helpers" which are only available in the storybook env (they won't be published as part of the phenotypes library):

1. a `<Box />` component that can be rendered within in a `<Canvas>` to demonstrate a utility class (or group of classes). These Boxes automatically get hover-tooltips with the className(s), so that we can see that without having to look through the "show code" content, if desired. The boxes also automatically get darker as they become more nested, so you can demonstrate how utilitlies between parents & children (& grandchildren, etc) work together. 
2. You can render a `<UtilityClassSuffixDocs />` block directly in the MDX to add a collapsible section that explains breakpoint/print/etc suffixes. That content will appear in many places, so at least we don't have to repeat the code for that everywhere.

color

overflow

text


styles/_features.scss
```diff
@@ -15,30 +15,17 @@ $enable-reboot: true !default;
 // Utility features
 // ------------------------------
 
-// Display utilities: 95 lines of CSS (approx 2kb)
```
NickDunkman, line 18: I don't want to maintain these line counts + post-build-file-sizes, so removing these comments

package.json
```diff
@@ -55,6 +55,7 @@
     "postcss": "^7.0.32",
     "postcss-preset-env": "^6.7.0",
     "react-addons-test-utils": "15.6.2",
+    "react-tooltip": "^4.2.21",
```
NickDunkman, line 58: Added this to dev dependencies for showing tooltips in docs -- may eventually replace with a Phenotypes Tooltip component if we ever add one of those.

src/utilities/color/color.stories.mdx
```diff
@@ -0,0 +1,186 @@
...
+#### Text colors
+
+<Canvas isColumn>
+  <div className="d-flex flex-wrap" style={{ gap: 10 }}>
+    <TextColorBox className="text-color-primary" hint="black, 0.86" />
+    <TextColorBox className="text-color-secondary" hint="black, 0.7" />
+    <TextColorBox className="text-color-hint" hint="black, 0.54" />
+  </div>
+  <div className="d-flex flex-wrap" style={{ gap: 10 }}>
+    <TextColorBox
```
NickDunkman, line 46: I considered sticking all these into some variables and mapping over them -- but end up just building out all the colors within the JSX like this so that the "show code" storybook feature provides helpful detail.

NickDunkman: (I could see eventually pulling/duplicating sass variable values into javascript, which would at least remove the need to list out the hex/rgba hints here)

src/utilities/color/color.stories.mdx
```diff
@@ -0,0 +1,186 @@
...
+    <BgColorBox className="bg-brand-accent" hint="#9e26bf" />
+    <BgColorBox className="bg-positive" hint="#00856f" />
+    <BgColorBox className="bg-negative" hint="#ab002b" />
+    <BgColorBox className="bg-warning" hint="#dd2c00" />
+    <BgColorBox className="bg-interactive" hint="#005eb0" />
+    <BgColorBox className="bg-error" hint="#ab002b" />
+  </div>
+</Canvas>
+
+## Non-themeable colors
```
NickDunkman, line 87: I used themeable / non-themeable as the top level section groupings in this doc, b/c I want people to get in the habit of using the themable-versions by default.

src/utilities/storyHelpers.tsx
```diff
@@ -0,0 +1,208 @@
...
+        borderRadius: 4,
+        padding: '5px 20px',
+      }}
+    >
+      <ToggledDocBlock
+        toggler={`About ${suffixType} suffixes in utility classes`}
+      >
+        {breakpoints && (
+          <Description>
+            `&#123;breakpoint&#125;` can be any valid breakpoint abbreviation:
```
NickDunkman, line 151: Storybook's `<Description>` component only accepts strings (markdown) via `children`, hence these escaped curly brackets 🤪

NickDunkman: P.S. -- using storybook's components in these doc block components so that the content gets styled to match storybook's docs styling.

styles/_reboot.scss
```diff
@@ -233,8 +237,7 @@ pre,
 code,
 kbd,
 samp {
-  font-family: monospace, monospace; // Correct the inheritance and scaling of font size in all browsers.
-  font-size: 1em; // Correct the odd `em` font sizing in all browsers.
+  @include text-monospace;
```
NickDunkman, line 240: I added a `.text-monospace` utility, so pulled these styles into that mixin.

styles/_theme_config.scss
```diff
@@ -11,12 +11,18 @@
...
   --font-family: #{inspect($font-family-base)};
-  --text-color-primary: #{$text-color-primary};
-  --text-color-secondary: #{$text-color-secondary};
-  --text-color-hint: #{$text-color-hint};
-  --text-color-primary-reversed: #{$text-color-reversed-primary};
-  --text-color-secondary-reversed: #{$text-color-reversed-secondary};
-  --text-color-hint-reversed: #{$text-color-reversed-hint};
+  --text-color-primary: rgba(#{extract-rgb($black)}, #{$alpha-800});
+  --text-color-secondary: rgba(#{extract-rgb($black)}, #{$alpha-700});
+  --text-color-hint: rgba(#{extract-rgb($black)}, #{$alpha-600});
```
NickDunkman, line 16: We have a few things that are using the `$text-color-primary|secondary|hint` vars (in both phenotypes & pare_web), such as the `<body>` text color(!), and those variables were _not_ themed. I changed things around so that the sass variables inherit from the css variables. (If you poke through the changes in the built css, you can see how the text is all themed properly now)

styles/_utilities_config.scss
```diff
@@ -0,0 +1,162 @@
...
+  $spacer-7,
+  $spacer-8,
+  $spacer-9,
+  $spacer-10,
+  $spacer-11,
+  $spacer-12,
+  $spacer-13
+) !default;
+
+// TODO: actually use these!
```
NickDunkman, line 52: pulled these over from `_variables.scss` -- we weren't actually using them, but I'll probably add `.width-full`/`.width-half`/etc when doing the sizing ticket.

styles/_variables.scss
```diff
@@ -15,42 +15,31 @@
 // ------------------------------
 // Colors
 // ------------------------------
-// Note: some of the original -400 and -500 colors were deprecated
-// as part of the colors refresh in October 2020. They are mapped
-// to the next available color for backwards compatibility.
```
NickDunkman, line 20: I don't see any reason to maintain these, I deleted all of them below.

pare_web is using a single one of these deleted colors to define a css class, but that css class isn't actually used.

styles/_variables.scss
```diff
@@ -327,13 +256,3 @@ $hr-reversed-border-color: rgba($white, $alpha-reversed-100);
...
-
-
-// ------------------------------
-// Other variables
-// ------------------------------
-// These are needed for Bootstrap reboot to work
-
-$table-cell-padding: .75rem !default;
-$text-muted: $text-color-secondary !default;
-$dt-font-weight: $font-weight-bold !default;
```
NickDunkman, line 339: pulled these over to `_reboot.scss`

src/utilities/color/color.stories.mdx
```diff
@@ -0,0 +1,186 @@
...
+
+## Themeable colors
+
+In general, you should use the [themeable](?path=/docs/guides-theming--page)
+color utilities, so that your app is themeable. The base theme’s color values
+are shown for convenience.
+
+#### Text colors
+
+<Canvas isColumn>
```
ahmedmodan, line 39: Can we render all of these color utlities without the source?

```suggestion
<Canvas isColumn withSource="none">
```
 
I don't see the utility in being able to view the source when it is only class names which are provided on hover and also viewing the source shows a react component that is nowhere to be found in any of our docs. What do you think?

src/utilities/color/color.stories.mdx
```diff
@@ -0,0 +1,186 @@
...
+    <TextColorBox className="text-color-interactive" hint="#005eb0" />
+    <TextColorBox className="text-color-error" hint="#ab002b" />
+  </div>
+</Canvas>
+
+#### Background colors
+
+<Canvas isColumn>
+  <div className="d-flex flex-wrap" style={{ gap: 10 }}>
+    <BgColorBox className="bg-brand-primary" hint="#630098" />
```
ahmedmodan, line 77: all of these wrapping divs for `BgColorBox` can have the `className` `text-color-primary-reversed` added to them. This way we can remove that `className` from the `BgColorBox` component. I don't think we need to show that class in the tooltip when we are trying to show what the class for background color is.

src/utilities/color/color.stories.mdx
```diff
@@ -0,0 +1,186 @@
...
+  </div>
+  <div className="d-flex flex-wrap" style={{ gap: 10 }}>
+    <BgColorBox className="bg-green" hint="#00856f" />
+    <BgColorBox className="bg-green-100" hint="#8af2d5" />
+    <BgColorBox className="bg-green-200" hint="#14c99c" />
+    <BgColorBox className="bg-green-300" hint="#00856f" />
+  </div>
+  <div className="d-flex flex-wrap" style={{ gap: 10 }}>
+    <BgColorBox className="bg-gray-100" hint="#f9f9f9" />
+    <BgColorBox className="bg-gray-200" hint="#eee" />
```
ahmedmodan, line 178: we should probably give `bg-gray-{100,200,300,400}` black text. It is not that readable with white text. We can probably get away with 300 and 400 with white text, but it reads easier with black

src/utilities/color/color.stories.mdx
```diff
@@ -0,0 +1,186 @@
...
+    <TextColorBox className="text-color-blue-400" hint="#002b66" />
+  </div>
+  <div className="d-flex flex-wrap" style={{ gap: 10 }}>
+    <TextColorBox className="text-color-green" hint="#00856f" />
+    <TextColorBox className="text-color-green-100" hint="#8af2d5" />
+    <TextColorBox className="text-color-green-200" hint="#14c99c" />
+    <TextColorBox className="text-color-green-300" hint="#00856f" />
+  </div>
+  <div className="d-flex flex-wrap" style={{ gap: 10 }}>
+    <TextColorBox className="text-color-gray-100" hint="#f9f9f9" />
```
ahmedmodan, line 130: The grays are pretty difficult to read here. especially 200 which almost seems like the same color as the background.

src/utilities/storyHelpers.tsx
```diff
@@ -0,0 +1,208 @@
...
+  className,
+  hint,
+  reversed = false,
+}: {
+  className: string;
+  hint: string;
+  reversed?: boolean;
+}) {
+  return (
+    <Box className={classes(className, reversed && 'bg-black')}>
```
ahmedmodan, line 180: can we hide this `bg-black` for the tooltip when it is used? Usually we are trying to display what the text color is and having `.bg-black` also pop up when it is only used to style the `Box` component seems like it should be unrelated. Maybe we can add a prop like `nonDisplayClassNames` or something like that. and those class names would not be added to the tooltips.

src/utilities/text/text.stories.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+<Canvas>
+  <Box className="text-weight-normal">.text-weight-normal</Box>
+  <Box className="text-weight-bold">.text-weight-bold</Box>
+</Canvas>
+
+## Style
+
+<Canvas>
+  <Box className="text-style-italic">.text-style-italic</Box>
```
ahmedmodan, line 102: can we add a little bit of margin or padding to the top of some of these canvases? The tooltip on some of these text canvases cuts off and the text is completely visibnle only when you hovering on the very bottm of the content.

ahmedmodan (commented):
Overall looks awesome! Thanks for redo-ing this! a few small suggestions (some grays are unreadable), and some nitpicks (tooltip spacing and not showing unrelated class names on hover.).

src/utilities/text/text.stories.mdx
NickDunkman: Yea. I'm gonna see if I can get rid of the `overflow: hidden` they're doing first ...

src/utilities/storyHelpers.tsx
NickDunkman: Yea I kinda went back and forth on that. I do like that removing `bg-black` in this case helps focus the docs, but also it might have people wondering if the other class is including black background styles? (in this case, probably no, but I could see that being an issue for many other situations).

To your point in the other comment, the `Box`/`BgColorBox`/`EtcBox` is hiding away some of the utility usage in a way that is opaque to the reader, so probably best to abandon the idea of providing full transparency into which utilities are being used to achieve the effect, and just focus in on the utility being documented.

ahmedmodan: Yeah in this case it doesn't necessarily make sense to show. But for cases where it does make sense, I can totally see how having two props, one for visible class names and one for hidden class names, would be helpful. I wonder if having the opposite would be more helpful. like `displayedClassName` would be all class names that are applied to the element and also visible on the tooltip. and then the regular `className` class would function like normal.

src/utilities/storyHelpers.tsx
```diff
@@ -0,0 +1,208 @@
...
+  const [tooltipId] = React.useState<string>(
+    boxContext.tooltipId || randomId()
+  );
+
+  const background =
+    boxContext.boxLevel >= BOX_LEVEL_COLORS.length
+      ? BOX_LEVEL_COLORS[-1]
+      : BOX_LEVEL_COLORS[boxContext.boxLevel];
+
+  const tooltipContent = className ? `.${className.replace(/\s+/g, ' .')}` : '';
```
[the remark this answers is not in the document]
NickDunkman: From that link:

Screen Shot 2022-03-02 at 13 36 39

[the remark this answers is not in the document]
NickDunkman: (that seems kinda silly)

src/utilities/text/text.stories.mdx
```diff
@@ -0,0 +1,193 @@
...
+  <Box className="text-underline">.text-underline</Box>
+  <Box className="text-line-through">.text-line-through</Box>
+  <Box className="text-monospace">.text-monospace</Box>
```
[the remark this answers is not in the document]
NickDunkman: Yea it's a little odd but it's actually really useful if we're using that pattern for the other classes -- I think ideally we'd update to something like:

```
.font-style-italic
.text-decoration-underline
.text-decoration-line-through
...etc
```

That's the way bootstrap does it -- different groups of these use a different style property (`font-style`, `text-decoration` ...). It's nice to be able to see which utilities can be used together and which are at odds. If you do something like this, you're not actually gonna see text with both an underline & a line-through:

```
<div className="text-underline text-line-through">lol</div>
```

src/utilities/color/color.stories.mdx
NickDunkman: https://github.com/aminohealth/phenotypes/pull/195/commits/9ba61f7db91d132c61eba619e3ebbf978dd3e0fd

Since we're not showing the source for these, I went ahead and pulled the color configs out into some data structures:

https://github.com/aminohealth/phenotypes/pull/195/commits/552628521a975a744ceedf57208f7e461c13d7f2

NickDunkman: ^ also fixed subtle grays in that last commit

NickDunkman: Screen Shot 2022-03-02 at 15 39 09

src/utilities/color/color.stories.mdx
NickDunkman: fixed -- see comment above

src/utilities/color/color.stories.mdx
NickDunkman: fixed -- see comment above

src/utilities/storyHelpers.tsx
NickDunkman: https://github.com/aminohealth/phenotypes/pull/195/commits/0ed250772f1d4edb2a608f0944bf439440f283ba

src/utilities/text/text.stories.mdx
NickDunkman: Fixed using a proper React Portal! https://github.com/aminohealth/phenotypes/pull/195/commits/ea9384d99e60a2b75a6eb88018a2c590d99e167f

NickDunkman: FYI: There's a little bit of weird behavior here as the tooltip butts up against the sidebar -- that's due to the wrapping iframe having an edge there.

ahmedmodan (approved):
tenor-146603561

PR #196, opened 2022-03-01
GUIDE-509: convert sizing utility docs to storybook

NickDunkman:
Preview is now working for this! Need to be on the VPN.

http://phenotypes.amino.io/?path=/docs/utilities-sizing--page

-----

I added in `min-width`, `max-width`, `min-height`, and `max-height`, as well as spacers `10`, `11`, `12`, and `13`. These add quite a bit to the compiled CSS, but we're adding many of them at the pare_web layer anyway (and where not, it's leads to a confusing developer experience).

I also added `-half`, `-third`, and `-fourth` (we already had `-full` in phenotypes). These are pretty commonly used in the web app to do things like "2/3/4 equally sized things side by side". I like calling using terminology like `fourth` instead of `quarter`, since it `fourth` has a smaller levenshtein distance from `four` 😎.


Makefile
```diff
@@ -15,10 +15,10 @@ clean:
 	docker-compose ${COMPOSE_ARGS} down
 
 shell:
-	docker-compose ${COMPOSE_ARGS} run --rm storybook-container bash
```
NickDunkman, line 18: updated the docker container name to match our general pattern per Jim's request

amino_build.json
```diff
@@ -2,13 +2,15 @@
...
     "deploy": {
         "preview": {
-            "cluster": "services-preview",
-            "region": "us-west-1",
-            "service": "phenotypes"
+            "cluster": "preprod-dmz",
+            "region": "us-west-2",
+            "service": "preview-phenotypes",
+            "profile": "aminotf",
+            "registry": "043538302950.dkr.ecr.us-west-2.amazonaws.com"
```
NickDunkman, line 13: I added deployment links to the bottom of `phenotypes-manual` builds in Jenkins -- these values I changed from were an older version of our AWS infrastructure.

jessiditocco (approved):
Looks good! Good call moving these sizers here. I feel like we were starting to add some of these in pare_web for the care dashboard ✨

PR #198, opened 2022-03-03
GUIDE-510: convert spacing utilities to storybook

NickDunkman:
Requesting feedback on this draft! Don't need line-by-line feedback, just wondering if people like the overall change.

I'd like to switch over to the way Bootstrap does their utilities: they have a [big data structure](https://github.com/twbs/bootstrap/blob/main/scss/_utilities.scss#L5) (sass map) that defines the configuration for all their utilities, and then they run through that [with some code](https://github.com/twbs/bootstrap/blob/main/scss/utilities/_api.scss) that is similar to the breakthrough loops in our existing utility files, and dump out all the css rules.

I think this is better for us because:

1. bootstrap decided to migrate from the way we do it (based on bootstrap 4) to this new "api" system for a reason
2. our utility declarations in the code are more readable (e.g. you can see a group of logic is for the "negative-margin-y" utility, it should have responsive classes, print classes, etc etc).
3. This would make the way we "extend" phenotypes utilities in pare_web a lot better. Right now we manually add new css rules, which are at least using variables like `$spacer-1`, but can be mis-named, leave many of the responsive versions undefined, etc. Making overrides to the utility config object would mean that things we add to pare_web (and then go populate all over the JSX codebase) are very likely exactly how we'll want to add them to phenotypes eventually.

Here's how adding/customizing in pare_web would work:

```scss
@import 'phenotypes/_variables.scss';

$spacer-14: 1billionpx;

$utilities: map-merge(
  $utilities,
  (
    // add a new utility!
    "border-size": (
      responsive: false,
      property: border-size,
      class: bs,
      values: $sizer-1 $sizer-2,
      print: true,
    ),
    // add a new modular scale number to margin utilities!
    "margin": map-merge(
      map-get($utilities, "margin"),
      (
        values: map-merge(
          map-deep-get($utilities, "margin", "values"),
          (14: $spacer-14),
        ),
      ),
    ),
  ),
);

@import 'phenotypes/_phenotypes.scss';
```

Of course, we could organize that all into some helpful file names:

```scss
@import 'phenotypes/_variables.scss';
@import 'phenotypes-extensions/border';
@import 'phenotypes-extensions/spacing';
@import 'phenotypes/_phenotypes.scss';
```

## How would the rollout work?

1. convert spacing to this utilities API system in this PR
2. circle back and convert other utilities as part of `typeybook` project
3. after that, there should be minimal changes in the built css (i did remove "negative padding" classes in this PR -- negative paddings aren't a real thing 🤣 )
4. Then we can pull this into pare_web when we release phenotypes v12, and can incrementally bring our phenotypes extensions into the new system (it doesn't break anything to continue to define phenotypes-esque classes "manually")

## Further reading

1. Bootstrap's docs on how to work with their utilities api: https://getbootstrap.com/docs/5.1/utilities/api/


styles/mixins/_utilities.scss
```diff
@@ -0,0 +1,124 @@
+// Copied from Bootstrap's mixin for converting utility API declarations to CSS rules
+// https://raw.githubusercontent.com/twbs/bootstrap/efc59146671bb671993353cc1f0a0daff1bfe554/scss/mixins/_utilities.scss
+//
+// ... with 2 tweaks:
+// 1. Changed $infix to $suffix (d-none-sm instead of d-sm-none)
+// 2. Added a `noDash` feature so we can keep doing mt5 instead of mt-5
+
+@mixin generate-utility($utility, $suffix, $is-rfs-media-query: false) {
```
NickDunkman, line 8: FYI read the comments I added above this function declaration -- this function is copied from bootstrap w/ some minor tweaks.

ahmedmodan:
i think overall this feels much more extendable than our current way of doing things. I do wonder about the overriding values. The nested `map-merge` and `map-gets` feels like it can be error prone especially for those that want to override but are not super well versed in sass utilities. Also **overriding** this way feels like you need more knowledge on the exact data structure in sass.  Perhaps for those who only just want one or two more spacers, a sass utility function that handles those deep merges and gets would be nice?

ahmedmodan:
Implementation also sounds pretty good to me. I would also prefer links to the bootstrap documentation on how to use the api within our own code. Or a link to a guide mirroring the bootstrap documentation in our own guides section.

NickDunkman:
> Or a link to a guide mirroring the bootstrap documentation in our own guides section.

Yea I was thinking I'd add a page to our own site!

NickDunkman:
> Also overriding this way feels like you need more knowledge on the exact data structure in sass. Perhaps for those who only just want one or two more spacers, a sass utility function that handles those deep merges and gets would be nice?

@ahmedmodan Yea -- I think in general we wouldn't see overrides happening this way. For to add spacers, you can just edit the `$spacers` sass var. But the option is there, if there's a good case for it. Much more common would be adding a new utility, which I could provide a sass function for out of the box. Something like this?

```scss
$utilities: add-utility($utilities, "my-utility", ( class: .... ));
```

ahmedmodan:
Oh nice,  i was not thinking about just modifying the `$spacers` var. That is a much easier solution. For adding another utility I think the way we have it now is probably fine. If you feel like an `add-utility` function would be better or easier that is cool with me, but a not absolutely necessary abstraction.

PR #199, opened 2022-03-04
GUIDE-512: sass hot reloading in storybook + setup /dist system

NickDunkman:
See GIF below to witness hot reloading!

Now that we no longer need to recompile CSS as part of the development workflow -- took the opportunity to clean up our `/dist` system, along with some rollup stuff.

---

### /dist

This directory houses our published JS/CSS, source maps, and now, a types file! I went back and forth on removing these from version control -- ultimately, it was too painful to see diffs to the CSS when making changes to SASS files, so I wanted to retain the option to easily rebuild CSS + see the diff from before your change. Keeping these files in version control makes that much simpler. It also enables people to grab the compiled files w/o going through NPM (e.g. our webflow public site does not use NPM to get the phenotyes CSS). The process for maintaining this directory will be as follows (I'll doc this in the README later):

1. At any time, you can run `make dist` to recompile the `/dist` directory.
  a. You can spot check the changes and not check them in (we will not require dist files to be updated in every PR)
  b. Or you can check them in if you think the PR reviewer will benefit from seeing them
2. Whenever we decide to publish a new version of phenotypes, we'll bump the version as well as run `make dist` and check in the changes, as part of the same PR. There may or may not be changes to dist files, depending on if people had already included some of the diffs in other PRs since the last version.
3. if you want to see compiled files, you can look at any version tag in Github & then navigate to the /dist folder.

---

### rollup changes

* We're now building the CSS in rollup! I removed the old `/scripts` system. You can see that the built CSS has zero changes (aside from adding a sourcemap reference).
* Added a CSS sourcemap
* Added a `dist/index.d.ts` types file! This only includes types for things that can be reached via the public API.
* I removed the `prepublishOnly` target -- when we run `npm publish`, it will just be publishing the `/dist` files that have already been packaged w/ that version bump in source control.

---

### misc

* We're only publishing an ESM version of our javascript library now (i removed the commonJs companion). It now lives at `dist/index.js` instead of `dist/phenotypes.esm.js`
* removed the non-themable version of our CSS
* renamed `/public_api.js` to `/index.ts`

---


package.json
```diff
@@ -89,8 +91,7 @@
     "url": "https://github.com/aminohealth/phenotypes/issues"
   },
   "homepage": "https://github.com/aminohealth/phenotypes#readme",
-  "module": "./dist/phenotypes.esm.js",
-  "main": "./dist/phenotypes.cjs.js",
+  "main": "./dist/index.js",
```
NickDunkman, line 94: I assume having an ESM-based file in the "main" option will work?

src/components/Slider/Slider.tsx
```diff
@@ -2,21 +2,21 @@ import * as React from 'react';
 import classes from 'classnames';
 import keycode from 'keycode';
 
-function getFillStyle(valueAsPercentage: number) {
+function getFillStyle(valueAsPercentage: number): React.CSSProperties {
```
NickDunkman, line 5: Had to fix some typescript errors in Slider.tsx in order for rollup to be able to build our JS.

configs/rollup.phenotypes-library.config.js
```diff
@@ -4,49 +4,89 @@ import babel from '@rollup/plugin-babel';
...
-    },
-    // output commonjs format for older workflows
-    {
+export default [
+  // 1st rollup step:
+  //   - first, clear out existing files in /dist
+  //   - compile phenotypes.css and write to /dist/css
+  //   - copy sailec.css & webfonts directory to /dist/css
+  {
+    input: path.resolve(WORKSPACE_ROOT, 'styles/phenotypes.scss.js'),
```
NickDunkman, line 22: It looks like we can't have rollup `input` an scss file here, so had to import the scss into a js file. (that's how `rollup-plugin-scss` recommends doing it)

.storybook/main.js
```diff
@@ -6,4 +8,13 @@ module.exports = {
   addons: ['@storybook/addon-links', '@storybook/addon-essentials'],
   staticDirs: ['./public'],
   framework: '@storybook/react',
+  webpackFinal: async (config) => {
```
ahmedmodan, line 11: `webpackFinalFinalV2`

configs/rollup.phenotypes-library.config.js
```diff
@@ -4,49 +4,89 @@ import babel from '@rollup/plugin-babel';
...
+  },
+  // 2nd rollup step:
+  //   - compile JS & write to /dist
+  //   - write ts declarations *per* each individual module to /dist/dts
+  {
+    input: path.resolve(WORKSPACE_ROOT, 'index.ts'),
+    output: {
       file: path.resolve(WORKSPACE_ROOT, packageJSON.main),
-      format: 'cjs',
+      format: 'esm',
```
ahmedmodan, line 51: 😍

package.json
ahmedmodan: Yeah But I would also be ok just leaving a module entry too. webpack by default resolves via the following order: `['browser', 'module', 'main']`
In our app, we target 'web' but that is the same as the default.
We can override this in our app though: https://webpack.js.org/configuration/resolve/#resolvemainfields

ahmedmodan (approved):

PR #200, opened 2022-03-17
GUIDE-516: add jest tests, tsc, eslint, and stylelint

NickDunkman:
- replaced mocha/enzyme/etc with `jest` & `testing-library`, rewrote all the tests & got them passing
- added `stylelint` -- we previously had some configuration for scss-lint, but stylelint seems to be the dominant style linter now. I have stylelint configured to ignore any issues that would conflict with prettier. As a result, there are many stylistic changes to sass files here.
- got `eslint` passing
- got `tsc` passing
- `make test` runs all of those -- but I also added individual make targets for each, if we want to have each be a separate check in github.
- added the jest addon to storybook -- see demo below!

**out of scope**

- making significant changes to Phenotypes components to make them more accessible / more "correctly" testable with test-library -- for this project, trying to limit the structural changes to our components
- I think I'll add a `prettier --check` tester in the future, but we need to run prettier on everything first.
- setting up github checks (will wait til we move this into the parelabs space)

---

### Demo of `make test` running without errors

Screen Shot 2022-03-20 at 16 43 54

---

### Hot reloading test results in the storybook jest addon

this shows running `make run` and `make jest_watch` in 2 separate terminals, and how changes to test results hot reload into storybook!

ezgif-5-699415c3bb


loganripplinger (changes requested):
1. Can you update the README with instructions?
    1. Using `make run` and `make jest_watch` together and how that is the preferred dev workflow now.
    2. Probably also worth explaining that `make test` now runs jest, tsc, and eslint.
3. When I run `make test` I'm getting 8000 errors from `eslint` that are all `no-mixed-spaces-and-tabs`. 😕

NickDunkman:
@loganripplinger I have a separate task to rewrite the README -- the entire thing is wrong now and needs to be rewritten to match all of the other changes I've been making in the `typeybook` branch. That will be done before `typeybook` is merged into `dev`.

https://linear.app/amino/issue/GUIDE-515/update-phenotypes-readme

For the eslint errors, those are probably in the compiled files in `/dist` and/or `/static` -- updated the `.eslintrc` to ignore those here: https://github.com/aminohealth/phenotypes/pull/200/commits/26d925e55e7109a333dac0637bc3aa62939f74d8

Any other errors are probably due to older fractal files that have been removed from source control -- If you still have errors, let's do a Zoom so I can debug?

NickDunkman:
In the meantime, since there isn't a proper README with instructions -- the correct way to run the entire test suite now is:

```bash
$ make build
$ make test
```

loganripplinger (approved):
Still happening; however I scrolled up and the source of all the errors is as you expected: `/usr/src/app/fractal_assets/js/bundle.js`. Want to add that to the eslintrc ignore?

Approving in the that that is what you expected and don't want to do that.

NickDunkman:
@loganripplinger Ah! I had removed that file locally. https://github.com/aminohealth/phenotypes/pull/200/commits/114ea53a79b20377b16196a7441c9f3a2c994c6d

PR #202, opened 2022-03-22
GUIDE-507: convert flexbox utilities to storybook + utilities API

NickDunkman:
http://phenotypes.amino.io/?path=/docs/utilities-flexbox--page

I completely rewrote our flexbox docs -- things are now arranged by the effect you might be trying to achieve, rather than the style properties used. I think it should make it a lot easier to understand how to use flexboxes.

Also, rewriting our utilities into the new utilities API is kinda fun!


src/utilities/spacing/_spacing.scss
```diff
@@ -36,187 +36,186 @@ $negative-spacers: ();
   }
 }
 
-$default-utilities: map.merge(
```
NickDunkman, line 39: this file is all moving things into an overridable `$default-spacing-utilities` sass var, + whitespace changes

src/utilities/flexbox/flexbox.stories.mdx
```diff
@@ -0,0 +1,478 @@
...
+      <Box className="p2">inline</Box>
+      <Box className="p2">flex</Box>
+      <Box className="p2">container</Box>
+    </Box>
+  </div>
+</Canvas>
+
+## Direction
+
+Every flexbox functions in one direction: either `row` (horizontal, the default)
```
jessiditocco, line 52: I think you may be missing a word here!
```suggestion
Every flexbox functions is in one direction: either `row` (horizontal, the default)
```

src/utilities/flexbox/flexbox.stories.mdx
```diff
@@ -0,0 +1,478 @@
...
+| **{value}** | effect                                                     |
+| ----------- | ---------------------------------------------------------- |
+| `start`     | items are aligned to the start of the flexbox              |
+| `end`       | items are aligned to the end of the flexbox                |
+| `center`    | items are aligned in the center of the flexbox             |
+| `baseline`  | items are aligned to the baseline                          |
+| `stretch`   | (default) items stretched equally to fill the full flexbox |
+
+<Canvas isColumn>
+  <Box ttClassName="d-flex align-items-start height-9">
```
jessiditocco, line 154: What is this `ttClassName`?

src/utilities/spacing/_spacing.scss
jessiditocco: I'm not sure how to test this, any suggestions?

src/utilities/flexbox/flexbox.stories.mdx
NickDunkman: I'm using "functions" as a verb there -- like "Jessi functions as a PR reviewer". I can prolly word that better 🤔

src/utilities/flexbox/flexbox.stories.mdx
NickDunkman: tt stands for toolTip -- any classes in there appear in a popup in the storybook UI when you hover over that `<Box />`:

Screen Shot 2022-03-23 at 15 40 23

src/utilities/flexbox/flexbox.stories.mdx
jessiditocco: Ohh I misread that! I think its fine as it is. I read that initially as "Every flexbox function in one direction"

src/utilities/spacing/_spacing.scss
NickDunkman: At this stage, I test in two ways:

- look at the changes to the`phenotypes.css` file & make sure that everything is as expected -- it's kind of a tedious process.
- look at is whether the demos on the storybook site look right -- unfortunately there aren't many visual demos of the spacing utilities: http://phenotypes.amino.io/?path=/docs/utilities-spacing--page

The ultimate best way to make sure this didn't break the spacing utilities is to look at what happens in the web application when we pull the new version of phenotypes in -- if the site looks all wacky, then there may be problems with the spacing utilities. I will be publishing a "release candidate" of phenotypes before publishing a new major version after this project, so will be able to look at that.

I don't anticipate there being any issues though -- the `phenotypes.css` file has all the same rules.

jessiditocco: I see! Okay -- I looked through all of the files here, and extra close at the `phenotypes.css` again this morning.

jessiditocco:
Looks ready to me! :shipit:

PR #203, opened 2022-03-24
Convert all utilities to use Utilities API, add docs

NickDunkman:
**Note: the diff for this is quite large, most of it is due to ordering changes in the built css file -- you can probably skip over that 🤷 . I checked that as I was going, and believe everything is in order. We'll do a dry-run of importing a release candidate of phenotypes into pare_web when this project is finished, and will be able to see if there are any issues with the css then.**

----

New docs for our API is are here: http://phenotypes.amino.io/?path=/docs/utilities-api--page

----

So one thing I ran into when pulling in all of our utilities into the API system -- we have some utilities that apply multiple style properties with different values, e.g.:

```css
.text-3 {
  font-size: 16px !important;
  line-height: 24px !important;
  letter-spacing: 0 !important;
  font-weight: 400 !important;
}
```

Bootstrap does not have any utility classes that do that -- so their system for generating the utility classes does not support it. Their `generate-utility` mixin is very complex, so in order to build support for this into our utility API:

1. I first removed a whole bunch of features that `generate-utility` was supporting, which we aren't currently using, and I don't really see us using ever (`state` could make sense in the future, OR we could just keep things simpler 🤷 ). Not doing this would have made step 2 a nightmare.
2. Then I added the ability to specify a "map of maps" for a particular utility, which gives you full control of the style properties. Docs/example usage here: http://phenotypes.amino.io/?path=/docs/utilities-api--page#as-a-map-of-maps-or-of-maps-and-lists

Circling back to the text sizing utilities -- I added functions for getting text style properties as a map, alongside our existing mixins. The function versions are then used when defining the utilities. Fun fact: you can use the same name for both a function and a mixin! E.g.:

```scss
// takes a SASS map and outputs as style properties to the buffer
@mixin unpack($style-mapping, $important: false) {
  @each $property, $value in $style-mapping {
    #{$property}: $value if($important, !important, null);
  }
}

// returns a SASS map
@function text(
  $font-size: $font-size-3,
  $line-height-ratio: $line-height-ratio-3,
  $letter-spacing: letter-spacing(0),
  $font-weight: $font-weight-normal
) {
  $line-height: line-height-px($font-size, $line-height-ratio);

  @return (
    font-size: $font-size,
    line-height: $line-height,
    letter-spacing: $letter-spacing,
    font-weight: $font-weight
  );
}

// returns a SASS map
@function text-1($font-weight: $font-weight-normal) {
  @return text(
    $font-size: $font-size-1,
    $line-height-ratio: $line-height-ratio-3,
    $letter-spacing: letter-spacing(-2),
    $font-weight: $font-weight
  );
}

// outputs style properties to the buffer
@mixin text-1($font-weight: $font-weight-normal, $important: false) {
  // this is calling the `text-1` @function above!
  @include unpack(text-1($font-weight), $important);
}
```


gelsto (approved):
Looks great! Thank you for the detailed PR description, it was very helpful 🙏

PR #204, opened 2022-03-26
Add ability to run storybook statically using nginx

NickDunkman:


ewdicus:
I haven't reviewed this in detail yet, but I'm curious why we wouldn't want to check-in the static site. It seems fairly common to do that, and makes it easy to separate out the nginx container (or to move to other static hosting like S3 or GitHub Pages if we later wanted to).

NickDunkman:
@ewdicus Basically every change to this repo will result in a change to the static build, and it's a pretty large amount of code, so every PR would basically be 2x the size it would otherwise be (plus, if multiple people were working on things, there'd be more opportunities for conflicts).

I think if we wanted to move to S3 later, it would be a trivial change (2 lines of code) to pull out the static-generation-only code into a separate make target. (I'm not sure about going the Github pages route, since we may wanna pull the site under our VPN eventually.)

ewdicus (commented):
This seems reasonable, but reading nginx conf always leaves me 😵. I'm going to pull this down and run locally as well. 

In the meantime, would be great to get someone from core-infra to review the nginx configuration just so we have another set of 👀 .

Makefile
```diff
@@ -1,79 +1,126 @@
...
+# run storybook in static mode
+# this runs an nginx server that serves from /storybook-static
 run_static:
-	docker-compose ${PROD_COMPOSE_ARGS} up
+ifdef IN_DOCKER_CONTAINER
+	npm run jest || true
+	npx build-storybook -o storybook-static
+	nginx -g "daemon off;"
+else
+	docker-compose ${COMPOSE_ARGS} run --rm -p 3001:80 storybook run_static
```
briankzimmer, line 56: Do the ports have to be different? I imagine we're only going to use one of the two `run`s at a time?

Dockerfile
```diff
@@ -2,14 +2,28 @@ FROM node:14.18.2-stretch-slim
...
+# install make, git, vim, nginx
+# (using nginx-extras because it includes perl module)
+RUN apt-get update
+RUN apt-get -y install build-essential git vim nginx-extras
+RUN apt-get clean
+
+# Configure NGINX
+RUN service nginx stop
+COPY ./configs/nginx.conf /etc/nginx
+RUN useradd --system --no-create-home --shell /bin/false nginx
```
briankzimmer, line 17: Generally we want to only use each Docker container for one thing, and installing nginx inside a node container feels like overloading this container to me.

I think this could also be leading to the bifurcation of the different use cases in the Makefile.

Would it be possible to stick with the initial goal of not checking in the static assets to the repo, but use a different Dockerfile structure so that we can only use this file for building/running/developing, and have a multi-stage Dockerfile that builds a static nginx container?

If we did that, I think we could remove all of the if statements from the Makefile?

briankzimmer: pseudo-ish code for the multi-stage build:
```dockerfile
FROM development-node-builder
npm run jest || true  # Needed?
npx build-storybook -o storybook-static

FROM nginx
COPY nginx.conf /etc/nginx
COPY --from=development-node-builder output-files nginx-dest-dir
```

...plus some other stuff to make it prod-worthy?

briankzimmer (commented):
Can comment more on the nginx config stuff specifically soon, but wanted to raise a more fundamental structure question first.

configs/nginx.conf
```diff
@@ -0,0 +1,86 @@
+user  nginx;
+worker_processes  2;
+env DISABLE_HTTPS;
+env NGINX_HOST_LIST;
```
briankzimmer, line 4: Before diving too deep into this config file, I feel like I'm missing some context. There's a lot of stuff going on here, and it's deviating a fair amount from the nginx config patterns we already have.

Was this a recommended sample config file from a specific source? Was there a specific reason why some parts of this are being included / is all of this required, or can we do some simplification?

NickDunkman: This is almost entirely based on a combination of these two files from pare_web:

https://github.com/parelabs/pare_web/blob/dev/nginx/conf/nginx.conf
https://github.com/parelabs/pare_web/blob/dev/nginx/conf/conf.d/flask.conf

(I combined them together instead of having `nginx.conf` include a seconary file)

Then, I removed a bunch of stuff that appeared to be unnecessary for the phenotypes nginx server, plus modified it to get the `server_name` from a `NGINX_HOST_LIST` environment variable, instead of having the build process write a change to the config file.

Dockerfile
NickDunkman: Thanks, I'll give that a try!

Makefile
NickDunkman: I found myself wanting to use a different port so I could sanity check that I was seeing the static nginx version of the site, rather than a lingering process still running the dynamic node version.

Makefile
```diff
@@ -1,79 +1,126 @@
...
 
-# run storybook in dynamic mode (js/sass/etc will be hot-reloaded)
+# run storybook in normal mode (js/sass/etc will be hot-reloaded)
 run:
-	docker-compose ${COMPOSE_ARGS} up
-
-# runs a server pointed at the statically-built storybook site
-# TODO: this isn't working yet, need to setup nginx?
+ifdef IN_DOCKER_CONTAINER
+	npm run jest || true
```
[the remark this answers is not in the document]
NickDunkman: Yea I'm doing this intentionally -- it's being run to generate the test results .json file, and if the tests fail, I still want to run the site, and show that there were test errors. I'll figure out how to make this better self documented.

Makefile
```diff
@@ -1,79 +1,126 @@
...
-# builds the public JS & CSS API of phenotypes (into ./dist)
+# compiles the public JS & CSS API of phenotypes (into ./dist)
 dist:
-	docker-compose ${COMPOSE_ARGS} run --rm --entrypoint='npm run' storybook buildDist
-
-# builds the static version of the storybook site (into ./static)
-static:
-	docker-compose ${COMPOSE_ARGS} run --rm --entrypoint='npm run' storybook buildStatic
+ifdef IN_DOCKER_CONTAINER
+	npx rollup --config ./configs/rollup.phenotypes-library.config.js
```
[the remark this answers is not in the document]
NickDunkman: Yea this made a little more sense to me when having things in a single container, if I get the multi-stage build / separate containers thing going, I'll simplify that back.

configs/nginx.conf
briankzimmer: lol I stand corrected! I really didn't remember seeing some of those things. My bad.

Let me dig into this stuff a bit more.

briankzimmer: Closing out this comment in favor of more granular ones in a forthcoming review comment set.

briankzimmer: Well, nevermind, I actually don't have a button to resolve this comment thread. Feel free to do so on your end, though.

configs/nginx.conf
```diff
@@ -0,0 +1,86 @@
...
+
+error_log /dev/stdout error;
+
+pid        /var/run/nginx.pid;
```
briankzimmer, line 8: I doubt we actually need to configure this -`/var/run/nginx.pid` is the default location for the file on Ubuntu, but I don't think there's an advantage to changing whatever the default is?

configs/nginx.conf
```diff
@@ -0,0 +1,86 @@
+user  nginx;
+worker_processes  2;
```
briankzimmer, line 2: If we're going to configure this to a specific number other than `1` or `auto` it probably makes sense to document the reasoning behind it. I'm assuming we're using `2` so that nginx can still service other requests if one process gets bogged down with a long running request. (e.g. doing an expensive gzip operation)

A quick thought on using `worker_process auto` to automatically select the number of processes based on the number of cores AWS is giving us - this would return the number of cores provided by the underlying EC2 instance that we're running on in ECS, but the CPU units assigned may only give us a fraction of a single vCPU, which translates to X% of a CPU in any given second. I'm not sure how the number of cores factors into that yet.

configs/nginx.conf
```diff
@@ -0,0 +1,86 @@
...
+
+        # compression
+        gzip              on;
+        gzip_buffers      16 8k;
+        gzip_comp_level   4;
+        gzip_http_version 1.0;
+        gzip_min_length   1280;
+        gzip_types        text/plain text/css application/x-javascript application/json text/xml application/xml application/xml+rss text/javascript;
+        gzip_vary         on;
+        gzip_proxied      any;
```
briankzimmer, line 79: I'm assuming its worthwhile from an engineer experience perspective to enable gzip, but I'm wondering if we actually need to configure all of these options, especially since this isn't end-user facing. Could we get away with only setting `gzip on`, `gzip_comp_level`, and `gzip_types *`?

Since we're not proxying anything, that directive isn't needed.

The `vary` thing is important in theory, but in practice I get the impression it's not really useful?

configs/nginx.conf
```diff
@@ -0,0 +1,86 @@
...
+
+    server {
+        listen 80;
+
+        server_name $nginx_host_list;
+        root /data/phenotypes/storybook-static/;
+
+        real_ip_header X-Forwarded-For;
+        set_real_ip_from 10.0.0.0/8;
+        real_ip_recursive on;
```
briankzimmer, line 69: In theory we can still use these directives, though I'm wondering how much we care about this for an internal tool.

configs/nginx.conf
```diff
@@ -0,0 +1,86 @@
...
+
+    access_log  /dev/stdout log_uuid;
+
+    client_body_in_file_only clean;
+    client_body_buffer_size 32K;
+    client_max_body_size 30M;
+
+    sendfile        on;
+
+    keepalive_timeout  65;
```
briankzimmer, line 35: The default is currently 75s, maybe it was 65s in the past - do we have a reason to customize this?

configs/nginx.conf
```diff
@@ -0,0 +1,86 @@
...
+
+    gzip  on;
+
+    # Security stuff
+    server_tokens off;
```
briankzimmer, line 40: How much do we care about not emitting nginx version info on error pages? I'm assuming we're almost always going to be running the latest version?

briankzimmer (commented):
@NickDunkman I ended up using this PR as an excuse to do a comprehensive review of how we're configuring nginx.

I don't want to block your PR from merging based on sorting all of this out. Feel free to use whatever suggestions you think makes sense, and we can continue sorting out the details after you push this out.

briankzimmer:
I just realized I forgot to put this in the last high-level comment, but wanted to acknowledge that a good number of the above line-level comments came out of a conversation that @kstrauser and I had on Tuesday.

configs/nginx.conf
```diff
@@ -0,0 +1,86 @@
...
+
+    log_format log_uuid '$remote_addr - $remote_user [$time_local] '
+                        '"$request" $status $body_bytes_sent '
+                        '"$http_referer" "$http_user_agent" "$http_x_forwarded_for" "$request_time" "$cookie_uuid"';
+
+    access_log  /dev/stdout log_uuid;
+
+    client_body_in_file_only clean;
+    client_body_buffer_size 32K;
+    client_max_body_size 30M;
```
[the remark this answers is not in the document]
NickDunkman: Cool -- should I remove all three of these lines, or just `client_max_body_size`?

configs/nginx.conf
```diff
@@ -0,0 +1,86 @@
...
+        gzip_comp_level   4;
+        gzip_http_version 1.0;
+        gzip_min_length   1280;
+        gzip_types        text/plain text/css application/x-javascript application/json text/xml application/xml application/xml+rss text/javascript;
+        gzip_vary         on;
+        gzip_proxied      any;
+
+        location / {
+            alias /data/phenotypes/storybook-static/;
+            expires max;
```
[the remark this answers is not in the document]
NickDunkman: Yea I'm not really sure -- storybook automatically changes version strings on the asset URLs when the assets change. But I guess this would also cache the initial HTML page as well?

configs/nginx.conf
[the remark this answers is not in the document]
briankzimmer: Yeah, all three can go

NickDunkman:
Thanks for the in-depth review @briankzimmer! Most of these things were copied from web without really understanding what they were doing or choosing options explicitly -- so generally don't have justifications for why those values are there. Will hopefully update a lot of this stuff tomorrow!

PR #205, opened 2022-03-29
GUIDE-508: convert grid docs to story format

NickDunkman:
I took more inspiration from Bootstrap & created a new navigation section in storybook for the `container` and `grid` docs, called "layout". Those system do not use the new Utility API (all other utilities do), and they are also a little different from the utilities in that they're meant to be used as a system, and shouldn't just be applied to arbitrary elements to affect their style.

The preview site is in use by another PR that is using nginx + ECS changes that are not compatible with this branch, so if you want to check these new pages out, you'll have to run them locally:

```
$ make build
$ make run
```

Screen Shot 2022-03-29 at 13 45 01

Screen Shot 2022-03-29 at 13 44 53


styles/_config.scss
```diff
@@ -1,6 +1,5 @@
 // Feature toggles
 $enable-reboot: true !default;
-$enable-grid-classes: true !default;
```
NickDunkman, line 3: I think we'll generally phase these $enable things out over time to simplify things. Our Amino codebases won't really need to opt out of anything, IMO.

src/helpers/storyHelpers.tsx
```diff
@@ -15,7 +15,9 @@ const BoxContext = React.createContext<{
   // doesn't have a tooltip. We set a tooltipId in the root-level Box, and then
   // pass it down to all the descendants.
   tooltipId: string | null;
-}>({ tooltipId: null });
+  // Each "generation" of Boxes within a root gets a new shade of gray
+  boxLevel: number;
```
NickDunkman, line 19: I added this `boxLevel` thing back so I can have the bottom-most box not use transparency (otherwise the new "grid background" shows through and makes things harder to see)

ahmedmodan (approved):
lgtm!

library/.prettierrc.json
```diff
@@ -1,11 +0,0 @@
...
-  "trailingComma": "all",
-  "overrides": [
-    {
-      "files": ["*.js", "*.jsx"],
-      "options": {
-        "singleQuote": true
-      }
-    }
-  ]
-}
```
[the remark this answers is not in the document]
NickDunkman: We got one at the root of the repo now! https://github.com/aminohealth/phenotypes/blob/typeybook/.prettierrc.json

PR #213, opened 2022-04-06
Add Textarea component

NickDunkman:
cc @jessiditocco 

http://phenotypes.amino.io/?path=/docs/components-textarea--empty

We got a design with a Phenotypes Textarea in it! (In the appointment edit description modal!)

- I implemented this based on the visual style of the TextInput (in the phenotypes UI kit, Sumul mentions that they should have the same exact style).
- I also implemented a version of Ahmed's auto-resize functionality! I think I was able to make it a bit simpler, and still seems to work well across browsers.
- Is compatible with Formik

Screen Shot 2022-04-05 at 17 30 45

Screen Shot 2022-04-05 at 17 47 57


src/components/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/**
+ * Textarea allows users to enter & edit multi-line text.
+ * @param props {@link TextareaProps}
+ */
+export function Textarea({
+  className,
+  disabled = false,
+  rows = 2,
```
NickDunkman, line 61: Browsers use 2 as the default as well.

configs/rollup.phenotypes-library.config.js
```diff
@@ -45,7 +45,7 @@ export default [
   //   - compile JS & write to /dist
   //   - write ts declarations *per* each individual module to /dist/dts
   {
-    input: path.resolve(WORKSPACE_ROOT, 'index.ts'),
+    input: path.resolve(WORKSPACE_ROOT, 'src/index.ts'),
```
NickDunkman, line 48: I moved the main `index.ts` file into `/src` -- when trying to import it from the root of the repo, the import resolver apparently gets confused about there being a package.json there with a `main` pointing at a file in the `/dist` directory. Moving that `index.ts` file avoids that altogether.

src/guides/12-formik.stories.mdx
```diff
@@ -1,6 +1,6 @@
 import { Canvas, Meta } from '@storybook/addon-docs';
 import * as formik from 'formik';
-import * as phenotypes from '../..';
+import * as phenotypes from '..';
```
NickDunkman, line 3: This is the thing that was confused about `/index.ts` vs. `/package.json->main`

src/components/Textarea/_textarea.scss
```diff
@@ -0,0 +1,127 @@
...
+    border-color: var(--error-color);
+  }
+
+  .Textarea__control:focus {
+    border-color: var(--error-color);
+    box-shadow: 0 0 11px rgba(var(--negative-color-rgb-values), $alpha-600);
+  }
+}
+
+@if $enable-responsive-form-components {
```
ahmedmodan, line 90: do we need this? Phenotypes will only be used in our app now, and we always have it enabled.

ahmedmodan (commented):
would recommend adding explicit auto resize functionality.

How does this work with the existing textarea with auto resize usage in the current app? (we have a workflow where you can test with yalc: https://www.notion.so/aminohealth/Create-local-Phenotypes-package-to-share-across-your-local-environment-e17ca7c7065241938ba260199497c091) In my mind this should be the replacement component and we should get rid of the one we have in pare_web, when this is merged. I would not want two competing components with the intention that we get rid of one of them one day since over time they will end up diverging in functionality, making replacement difficult.

In firefox this shows one extra row than the number of rows. I think for this reason in my original implementation I manually set the height which was computed by line height * number of lines, instead of using the rows attr.

https://user-images.githubusercontent.com/13989876/162068991-76d2bd57-673b-4b18-910b-6880cdc00438.mp4

src/components/Textarea/_textarea.scss
NickDunkman: 🔥 https://github.com/aminohealth/phenotypes/pull/213/commits/77a702ff74314284670f794d7575472474b2a6de

src/components/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,159 @@
...
+  const theValue = (isControlledComponent ? value : uncontrolledValue) || '';
+
+  function handleChange(event: React.ChangeEvent<HTMLTextAreaElement>) {
+    !isControlledComponent && setUncontrolledValue(event.target.value);
+    onChange?.(event);
+  }
+
+  // If auto-resizing is enabled, compute the number of rows to use on the
+  // visible textarea
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
NickDunkman: Nice, TIL! https://github.com/aminohealth/phenotypes/pull/213/commits/26691b83d5758e398e88599789ae62875560f8a1

src/components/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,159 @@
...
+   */
+  disabled?: boolean;
+  /** The number of rows of text tall the Textarea should be. Adding more than
+   * that many rows of text will cause the Textarea to become scrollable
+   */
+  rows?: number;
+  /** If provided, the textarea will start at `rows` tall, and then grow
+   * up to `maxRows` tall as the user adds more text
+   */
+  maxRows?: number;
```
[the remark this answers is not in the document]
NickDunkman: Added an `autoResize` prop & set the default for `maxRows` to `Inifinity`

NickDunkman:
@ahmedmodan Updated to set a height instead of a `rows` prop as discussed offline. Is slightly different than your version in web since there are vertical padding & borders to consider.

NickDunkman:
(also discussed offline: not going to try to use this new component in the chat message textarea in pare_web due to the built in submit button -- will just namespace what we have in web under the chat message feature)

ahmedmodan (approved):
tenor-37691618

